PR #2, opened 2019-05-05
Set up a multistage build

fabianvf:
I'm starting to look at mc-router internals and figured I may as well PR whatever changes I'm making to ease my development. I prefer to let my builds happen in the containers so I can't mess it up environmentally, but if you prefer to leave it how it is I understand that too.


Dockerfile
```diff
@@ -1,3 +1,17 @@
+FROM golang:1.11 AS builder
+
+ENV CGO_ENABLED=0
+
+WORKDIR /src
+
+COPY go.mod /src/go.mod
+RUN go mod download
+
+COPY src /src/
```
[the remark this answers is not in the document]
fabianvf: Yeah, `src` is definitely wrong. I think the normal pattern is to have `cmd` and `pkg`

fabianvf:
I can still avoid the move if you want, though the reason I moved the libraries under `pkg` is it makes it much easier to handle the `COPY` of the src directories. I usually try to avoid `COPY`ing `.` so that my builds are faster, and it's also nice to not need a layer per library directory.

Dockerfile.build
```diff
@@ -0,0 +1,18 @@
+FROM golang:1.12 AS builder
+
+ENV CGO_ENABLED=0
+
+WORKDIR /build
+
+COPY go.mod /build/go.mod
+RUN go mod download
+
+COPY cmd /build/cmd
```
itzg: It'll look a little silly but have used

````
COPY . .
```

in cases like this to stage the whole project source after doing the dependencies download stage.

I'm not sure if I like the `pkg` convention either since that forces an extra `pkg` level into the module path of `github.com/itzg/mc-router/pkg/server` as an example in this case.

fabianvf: My issue with `COPY .` is that it invalidates the cache if anything changes in that directory, even if it has nothing to do with the actual build. As a dumb example, if I have my Dockerfile:
```
FROM golang:1.12 AS builder

COPY . /build
RUN go build -o mc-router /build/cmd/mc-router/main.go

FROM scratch

COPY --from=builder /build/mc-router /bin/mc-router

ENTRYPOINT ["/bin/mc-router"]
```
and I want to add an environment variable to my container, like so:
```
FROM golang:1.12 AS builder

COPY . /build
RUN go build -o mc-router /build/cmd/mc-router/main.go

FROM scratch

COPY --from=builder /build/mc-router /bin/mc-router
ENV EXAMPLE=true

ENTRYPOINT ["/bin/mc-router"]
```

Rather than properly caching and starting at the second to last step in the second container, it will go all the way back to my build container and start over from the `COPY .` task.

All that being said it's a fairly minor issue especially for a project of this size, just wanted to explain my reasoning before I reverted it. If you still prefer the top-level style rather than the `pkg` style I can move them back.

itzg: First of all, thank you for entertaining my questioning. Like you said, for a project this size it's not really a big deal either way.

however, I'm still leaning towards the top-level style and the original `Dockerfile` because I have been refining and using the same CI work flow with goreleaser and CircleCI over several projects. 

I'd still like the sound of an adjacent, development-time `Dockerfile.build` or similar and would clarify my suggestion to have that include the section:

```
WORKDIR /build
COPY go.mod go.sum .
RUN go mod download
COPY . .
RUN go build -o mc-router cmd/mc-router/main.go
```

It doesn't solve all the cache invalidation scenarios, but will at least cache the module downloading between iterations.

itzg: Sorry, I just noticed you did put this in `Dockerfile.build`, but the two step COPY still applies :)

sfxworks:
While you're here, consider going daemonless?

Example from https://github.com/GoogleContainerTools/distroless
```
# Start by building the application.
FROM golang:1.13-buster as build

WORKDIR /go/src/app
ADD . /go/src/app

RUN go get -d -v ./...

RUN go build -o /go/bin/app

# Now copy it into our base image.
FROM gcr.io/distroless/base-debian10
COPY --from=build /go/bin/app /
CMD ["/app"]
```

itzg:
> While you're here, consider going distroless?

A distroless debian (gcr.io/distroless/base-debian10) seems contradictory :) FYI, the goreleaser built binary is all the way distroless

https://github.com/itzg/mc-router/blob/c49099f1cf16a385f710128d0c9b4bf82aff10d3/Dockerfile#L1

sfxworks:
Oh
```
kubectl exec -it mc-router-8574c6f858-d96t4 -- /bin/sh
ERRO[0000] exec failed: container_linux.go:349: starting container process caused "exec: \"/bin/sh\": stat /bin/sh: no such file or directory"
exec failed: container_linux.go:349: starting container process caused "exec: \"/bin/sh\": stat /bin/sh: no such file or directory"
command terminated with exit code 1
```
Oh nice! Horay for minimal containers!

itzg:
Yep, static Go builds + scratch images for the security win.

vorburger:
@itzg close this old one now, as I just noticed that you seem to have done what this was about via ca55bde751962557bdb0e2f62da22c46d180409e ?

itzg:
Yes, thanks

PR #3, opened 2019-05-06
[wip] Inspect frames for commands

fabianvf:
Just testing out how tough it would be to build the server listing commands and stuff


itzg:
I'm excited to see what you're doing here and was also starting to think that at least sniffing the frames would be more doable than I originally feared. I'll try to look at this later today/tomorrow.

fabianvf:
Yeah, I think I just hit the wall of handling encryption, I think we'd need to MITM it.

fabianvf:
It looks gross but not undoable: https://wiki.vg/Protocol_Encryption

fabianvf:
I suppose we also have the option of retrieving the real private key from the server

itzg:
Oh yeah, it was the encryption that was the show stopper before when I considered going deeper. I'm guessing that's partly why bungeecord recommends configuring the backend minecraft servers in offline mode; however, that's a total guess.

itzg:
...speaking of bungeecord, maybe we should look at how they handled that aspect. Maybe it is more MITM by generating its own server key, yet talks to the individual backends via their respective encrypted channel.

molecule/default/asserts.yml
```diff
@@ -0,0 +1,44 @@
+---
```
itzg: Seeing the molecule files I realize the motivation for shifting the source code into a `pkg` directory.

fabianvf: yeah that's a really good point, if you're writing tests, updates to the tests may end up triggering rebuilds in the container, even if none of the application logic changes. For the molecule stuff, where I'm building the image in a docker-in-docker fashion (so it's available to the k8s cluster in the container without needing to be pushed to an external registry), `COPY .` adds significant overhead to the testing loop since it'll basically always result in a container rebuild, and by extension a cleanup and redeploy of the router deployment. 

On another note, how do you feel about the molecule stuff? I like it for end-to-end testing in CI environments, and I like using Ansible for basic API interaction with Kubernetes, but I'm also a contributor to molecule and also the maintainer of the k8s/ansible integration so I'm about as biased as possible. I can remove it from the final PR once it's working if you prefer a different testing infrastructure.

fabianvf:
So I think first step is getting this working in servers running in offline mode (no encryption it seems).

molecule/default/asserts.yml
itzg: I hadn't heard of molecule before, but I'm liking the look of it. I was already a fan of Ansible, so anything that plugs into that ecosystem seems good to me.

ryantheleach:
If you do end up investigating encryption, you may be interested in Velocity, I believe they have native decoding for better performance. https://github.com/VelocityPowered/Velocity/tree/master/native

PR #5, opened 2019-06-25
Make server address not case sensitive

JLSchuler99:
Store all addresses in lowercase. and check incoming connections as lowercase.
I had some users run into this issue, and can't think of a reason to have case sensitive hostnames for servers, but let me know if you have a different view.


itzg:
There is now a 1.5.0 release and docker image for this.

PR #20, opened 2021-11-08
add link to related project https://github.com/haveachin/infrared

vorburger:
I don't know if you would consider linking to a project by @haveachin in roughly the same space as this? Even if you don't seem to serve exactly the same use case, it might still be of interest to other folks looking at this. (Personally I'm considering to explore perhaps using BOTH of these projects together - one single `mc-router` as front-end entry point to dispatch, and multiple N `infrared` back-end "wrapping" instances for on-demand starting of Minecraft server containers via their respective StatefulSets. FYI @edewit.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,3 +171,7 @@ docker run -it --rm \
   goreleaser/goreleaser \
   release --snapshot --rm-dist
 ```
+
+# Related Projects
+
+* https://github.com/haveachin/infrared


itzg:
Wow, 1.7 million pulls in 7 months. It seems to be vastly more popular than mine. It looks like it covers mc-router's use cases, so I'm not sure why you'd need to include mc-router.

haveachin:
Thanks, but I'm actually not sure how it got so many Docker pulls. As for this PR, I think it would be cool to link similar projects even if one seems like a superset of the other.

itzg (commented):
If the link from infrared is merged in then I'm good with merging this one.

PR #22, opened 2021-11-12
feat(source ip): implement PROXY protocol for source IPs

Gabitzuu:
Might not be the best way to do things, but it works. Hopefully it'll get merged.


server/connector.go
```diff
@@ -189,6 +191,26 @@ func (c *connectorImpl) findAndConnectBackend(ctx context.Context, frontendConn
...
+		Command:           proxyproto.PROXY,
+		TransportProtocol: proxyproto.TCPv4,
+		SourceAddr:        clientAddr,
+		DestinationAddr:   remoteAddr,
+	}
+
+	_, err = header.WriteTo(backendConn)
+	if err != nil {
+		// handle error
+	}
```
itzg, line 212: Add a new command-line/env field to

https://github.com/itzg/mc-router/blob/8b54fdfb863bd2af7bbfb4fc38a750b311bdf34b/cmd/mc-router/main.go#L31-L44

such as `UseProxyProtocol` and then wrap this in a conditional block. The default would be false, keeping the behavior to existing users of this.

server/connector.go
```diff
@@ -189,6 +191,26 @@ func (c *connectorImpl) findAndConnectBackend(ctx context.Context, frontendConn
 	c.metrics.ActiveConnections.Add(1)
 	defer c.metrics.ActiveConnections.Add(-1)
 
+	// PROXY protocol implementation
+
+	remoteAddr, err := net.ResolveIPAddr("ip", strings.Split(backendHostPort, ":")[0])
```
itzg, line 196: You can actually use https://pkg.go.dev/net@go1.17.3#SplitHostPort to split that apart and grab the hostname, such as

```suggestion
      remoteHost, _, _ := net.SplitHostPort(backendHostPort)
	remoteAddr, err := net.ResolveIPAddr("ip", remoteHost)
```

server/connector.go
```diff
@@ -189,6 +191,26 @@ func (c *connectorImpl) findAndConnectBackend(ctx context.Context, frontendConn
...
+		Version:           2,
+		Command:           proxyproto.PROXY,
+		TransportProtocol: proxyproto.TCPv4,
+		SourceAddr:        clientAddr,
+		DestinationAddr:   remoteAddr,
+	}
+
+	_, err = header.WriteTo(backendConn)
+	if err != nil {
+		// handle error
```
itzg, line 211: Borrow from one of the blocks below and do something like:

```suggestion
		logrus.
			WithError(err).
			WithField("client", clientAddr).
			Error("Failed to write proxy header")
		c.metrics.Errors.With("type", "write").Add(1)
		_ := backendConn.Close()
		return
```

itzg (changes requested):
This is really good. Just a few changes and I'll be glad to merge this.

server/connector.go
itzg: FYI, documentation on that flags syntax:

https://pkg.go.dev/github.com/itzg/go-flagsfiller#section-documentation

Gabitzuu: Tbh, I literally forgot to do that, in my mind I said that I should do that 😅.

I'll do in a few hours, when I'll be at my PC.

Gabitzuu: implemented this in commit #6f4272d :D

Gabitzuu:
Finished all change requests. Feel free to check the code and merge if wanted.

cmd/mc-router/main.go
```diff
@@ -40,6 +40,7 @@ type Config struct {
 	InKubeCluster        bool     `usage:"Use in-cluster kubernetes config"`
 	KubeConfig           string   `usage:"The path to a kubernetes configuration file"`
 	MetricsBackend       string   `default:"discard" usage:"Backend to use for metrics exposure/publishing: discard,expvar,influxdb"`
+	SendProxyProtocol    bool     `default:"false" usage:"Send PROXY protocol to backend servers"`
```
itzg, line 43: Nitpick: I would prefer `UseProxyProtocol`

itzg (changes requested):
Thanks. One last minor wording tweak and this is good to go.

cmd/mc-router/main.go
itzg: BTW the pickiness here is only because this is the "public interface" of the app

Gabitzuu: I thought SendProxyProtocol would be better as I was thinking to add recieve as well

Gabitzuu: solved 👍🏼

itzg: > I thought SendProxyProtocol would be better as I was thinking to add recieve as well

That logic makes sense. In any case, can still use "receive" or "accept" terminology if/when needed.

itzg (approved):
Awesome

itzg:
Finally got the goreleaser config sorted out:

https://github.com/itzg/mc-router/releases/tag/1.13.0

server/connector.go
```diff
@@ -189,12 +193,45 @@ func (c *connectorImpl) findAndConnectBackend(ctx context.Context, frontendConn
...
+		header := &proxyproto.Header{
+			Version:           2,
+			Command:           proxyproto.PROXY,
+			TransportProtocol: proxyproto.TCPv4,
+			SourceAddr: &net.TCPAddr{
+				IP:   net.ParseIP(sourceAddrStr),
+				Port: sourcePort,
+			},
+			DestinationAddr: &net.TCPAddr{
+				IP: net.ParseIP(remoteHostStr),
```
itzg, line 211: @Gabitzuu a user is reporting a problem with this feature

```
level=info msg="Got connection" client="xxx.xxx.xxx.xxx:48064"
level=info msg="Connecting to backend" backendHostPort="velocity:25577" client="xxx.xxx.xxx.xxx:48064" server=xxx
level=error msg="Failed to write PROXY header" clientAddr="xxx.xxx.xxx.xxx:48064" destAddr=":0" error="proxyproto: invalid address"
```

and I'm now wondering why a hostname, "velocity" in this case, is being parsed as an IP address. Did this feature assume IP addresses were manually configured as the backend routes?

Gabitzuu: Hi and sorry for the late response. It's been a long time since I've wrote this implementation. I think that's what I've intended, to only use IP addresses but not hostnames.
I was a Golang noobie back then so this feature is not written in the best way.

itzg: Thanks for getting to me. That's totally fine and just wanted to check I wasn't missing something. I'll change this to use either hostname or IP address.

PR #26, opened 2021-11-20
Remove un-used option -kube-discovery (fixes #23)

### LINKED ISSUE #23: Missing doc for kube-discovery (or remove, if unused?)
What does the kube-discovery config do?

It's mentioned in the README, but not used in the Kubernetes example.

https://github.com/itzg/mc-router/search?q=kube-discovery does not show it actually being used in source code. Is it perhaps something left over that could be removed from the README?

vorburger:
fixes #23


itzg (approved):
Thanks

PR #28, opened 2021-11-20
Very minor clean-ups (import re-ordering by automated tooling)

vorburger:


vorburger:
This is also in #29, but is super trivial to review and can easily be merged already, wheres #29 may or may not have more feedback.

server/connector.go
```diff
@@ -3,15 +3,16 @@ package server
...
+
 	"github.com/go-kit/kit/metrics"
 	"github.com/itzg/mc-router/mcproto"
 	"github.com/juju/ratelimit"
 	"github.com/pires/go-proxyproto"
 	"github.com/sirupsen/logrus"
-	"io"
-	"net"
-	"strconv"
-	"time"
```
itzg, line 14: I'm fine with the changes, but worried our tools will fight each other. I'm using gofmt (via IntelliJ). What formatting tool did you use here?

vorburger: Agreed, we should definitely standardize (see also e.g. my own https://github.com/vorburger/MariaDB4j/issues/264 about basically the same in another project).  I use Visual Studio Code (VSC), which kept insisting on formatting the imports like this, which is why I raised this PR to trigger this discussion.

According to https://github.com/Microsoft/vscode-go/issues/230, VSC's default is `goimports` instead of `gofmt`, and indeed e.g. `goimports -w server/connector.go` does the same as what this proposed. Reading up a bit more about this, that issue 230 is fairly old, the latest doc for VSC about this seems to be https://github.com/golang/vscode-go/blob/master/docs/tools.md, and that stuff about `gopls` vs the _legacy mode_ is applicable here. I see a ⚡ after _Go_ in my status bar, which apparently means that I'm using `gopls` and not that _legacy mode_. That explains why a ` .vscode/settings.json` with a `"go.formatTool": "gofmt"` (alone) does not do the trick. And https://github.com/microsoft/vscode-go/issues/3059 has more about this, with a setting that simply disables organizing imports (which is a shame), and background about why this new style appears to be _"idiomatic go"_ and the now recommended style. That issue does not mention any possible way to make `gopls` format like `gofmt`. So I guess we have 2 choices here:

a) I disable organizing imports on my end. Not ideal, because you'll likely reformat my imports. That would be PR #33.

b) You adopt this (apparently) new _"idiomatic"_ style, and we (you or with PR from me) reformat all imports based on it.

I think b) would be better if you can get your IntelliJ to format it like this, but if that's a PITA, then let's do a). WDYT?

itzg: Yeah, I have been blissfully unaware of the whole gofmt, goimports, goreturns variants. I can configure IntelliJ for goimports, so I'm good with option b.

itzg:
Merge conflicts need to be fixed.

vorburger:
@itzg merge conflicts are fixed, and this now includes import reformatting for several files (with a separate commit from a few other very trivial changes), as per discussion above and option b.

itzg (approved):
Thanks and sorry for the delays getting back to you.

Issue #21, opened 2021-11-09
Sleeping Start enhancement 


sfxworks:
Similar to https://github.com/vincss/mcsleepingserverstarter

Since this router is also aware of service definitions, and service definitions contain references to deployment or statefulset object via label, it would be cool if the router could pick up and hold incoming connections and automatically scale the stateful set up from 0 to 1 and 1 to 0 when it knows that there have been no active connections for some time.

sfxworks:
Or maybe a better pattern would be to wrap that project around https://github.com/itzg/docker-minecraft-server and replace the `minecraftCommand` with your wrapper script?

itzg:
Yeah, I really like the idea of mc-router influencing the replica count of a server's statefulset. My only concern would be the amount of time in between and attempted connection and when the destination server would be up and ready to serve connections.

sfxworks:
To combat startup time sleeping start server just says "come back later" to the client until it's up. Also on a ceph NVMe cluster I've seen startup times as fast as 5 seconds for vanilla

vorburger:
I'm exploring this in https://github.com/itzg/mc-router/compare/master...vorburger:issue-11_autoscale-2 and much to my own huge surprise 😈  actually do seem to be making reasonable progress towards teaching mc-router this trick, to be able to "auto-scale" the replica count of a server's statefulset...

vorburger:
I now have a a working MVP on the branch linked above, and will that contribute it as a PR, once I've further cleaned up the code. But as was suspected above, the "user experience" is not really all that great (yet, with my initial PR) - because you at first get disconnected, and then have to manually re-connect a moment later, once the server has actually spun up and is ready to accept connections... it works, and sure beats manually having to `k scale statefulset test --replicas=1` - but it would of course be much nicer if the connection was kept "hanging", until the server is ready. But rome wasn't built in one day, so let's chip away at this step by step. Perhaps someone else wants to later implement such an improvment, on top of my (upcoming) basic PR for this.

itzg:
Implemented in #488

PR #29, opened 2021-11-20
First version of Auto Scaler

vorburger:
#21

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,8 +177,11 @@ kind: ClusterRole
 metadata:
   name: services-watcher
 rules:
-- apiGroups: ["", "apps"]
-  resources: ["services", "statefulsets", "statefulsets/scale"]
+- apiGroups: [""]
+  resources: ["services"]
+  verbs: ["watch","list"]
+- apiGroups: ["apps"]
+  resources: ["statefulsets", "statefulsets/scale"]
   verbs: ["watch","list","get","update"]
 ```
 
--- a/README.md
+++ b/README.md
@@ -61,9 +61,9 @@ Routes Minecraft client connections to backend servers based upon the requested 
     "backend": "HOST:PORT"
   }
   ```
-  
+
 * `POST /defaultRoute` (with `Content-Type: application/json`)
-  
+
   Registers a default route to the given backend. JSON body is structured as:
   ```json
   {
@@ -74,7 +74,7 @@ Routes Minecraft client connections to backend servers based upon the requested 
 * `DELETE /routes/{serverAddress}`
 
   Deletes an existing route for the given `serverAddress`
-  
+
 # Docker Multi-Architecture Image
 
 The [multi-architecture image published at Docker Hub](https://hub.docker.com/repository/docker/itzg/mc-router) supports amd64, arm64, and arm32v6 (i.e. RaspberryPi).
@@ -86,25 +86,25 @@ configures two Minecraft server services named `vanilla` and `forge`, which also
 network aliases. _Notice those services don't need their ports exposed since the internal
 networking allows for the inter-container access._
 
-The `router` service is only one of the services that needs to exposed on the external 
+The `router` service is only one of the services that needs to exposed on the external
 network. The `--mapping` declares how the hostname users will enter into their Minecraft client
 will map to the internal services.
 
 ![](docs/compose-diagram.png)
-  
+
 To test out this example, I added these two entries to my "hosts" file:
 
 ```
 127.0.0.1 vanilla.example.com
 127.0.0.1 forge.example.com
 ```
-  
+
 # Kubernetes Usage
 
 ## Using kubernetes service auto-discovery
 
 When running `mc-router` as a kubernetes pod and you pass the `--in-kube-cluster` command-line argument, then
-it will automatically watch for any services annotated with 
+it will automatically watch for any services annotated with
 - `mc-router.itzg.me/externalServerName` : The value of the annotation will be registered as the external hostname Minecraft clients would used to connect to the
    routed service. The service's clusterIP and target port are used as the routed backend. You can use more hostnames by splitting them with comma.
 - `mc-router.itzg.me/defaultServer` : The service's clusterIP and target port are used as the default if
@@ -142,11 +142,11 @@ metadata:
 
 ## Example kubernetes deployment
 
-[This example deployment](docs/k8s-example-auto.yaml) 
+[This example deployment](docs/k8s-example-auto.yaml)
 * Declares an `mc-router` service that exposes a node port 25565
 * Declares a service account with access to watch and list services
 * Declares `--in-kube-cluster` in the `mc-router` container arguments
-* Two "backend" Minecraft servers are declared each with an 
+* Two "backend" Minecraft servers are declared each with an
   `"mc-router.itzg.me/externalServerName"` annotation that declares their external server name(s)
 
 ```bash
@@ -157,8 +157,8 @@ kubectl apply -f https://raw.githubusercontent.com/itzg/mc-router/master/docs/k8
 
 #### Notes
 * This deployment assumes two persistent volume claims: `mc-stable` and `mc-snapshot`
-* I extended the allowed node port range by adding `--service-node-port-range=25000-32767` 
-  to `/etc/kubernetes/manifests/kube-apiserver.yaml` 
+* I extended the allowed node port range by adding `--service-node-port-range=25000-32767`
+  to `/etc/kubernetes/manifests/kube-apiserver.yaml`
 
 # Development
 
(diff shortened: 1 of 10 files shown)


server/k8s.go
```diff
@@ -1,21 +1,27 @@
...
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/clientcmd"
-	"net"
-	"strconv"
 )
 
 const (
 	AnnotationExternalServerName = "mc-router.itzg.me/externalServerName"
-	AnnotationDefaultServer      = "mc-router.itzg.me/defaultServer"
+	AnnotationDefaultServer      = "mc-router.itzg.meN/defaultServer"
```
vorburger, line 24: Oups, not sure what typo happened here - I'll undo this...

server/k8s.go
```diff
@@ -110,9 +120,56 @@ func (w *k8sWatcherImpl) startWithLoadedConfig(config *rest.Config) error {
...
+			},
+		},
+	)
+
 	w.stop = make(chan struct{}, 1)
-	logrus.Info("Monitoring kubernetes for minecraft services")
-	go controller.Run(w.stop)
+	logrus.Info("Monitoring Kubernetes for Minecraft services")
+	go serviceController.Run(w.stop)
+	go statefulSetController.Run(w.stop)
```
vorburger, line 172: I'm not sure this is correct, I'm still learning about goroutines.

itzg: This is right; however, I should have used `close(w.stop)` in the `Stop` function instead of the following to allow for multiple "listeners":

```
		w.stop <- struct{}{}
```

server/routes.go
```diff
@@ -125,24 +135,24 @@ func (r *routesImpl) SetDefaultRoute(backend string) {
...
-		return r.defaultRoute, address
-	} else {
-
-		if route, exists := r.mappings[address]; exists {
-			return route, address
-		} else {
-			return r.defaultRoute, address
+	if r.mappings != nil {
+		if mapping, exists := r.mappings[address]; exists {
+			if err := mapping.waker(ctx); err == nil {
```
itzg, line 148: I would prefer that the action of waking the backend happen outside of this function since a "find" function really shouldn't have any side effects. I'd say have this function return `(string, string, func(ctx context.Context) error)` and then have `connectorImpl.findAndConnectBackend` call it.

server/k8s.go
```diff
@@ -154,6 +212,45 @@ func buildDetails(service *v1.Service, externalServiceName string) *routableServ
...
+		if statefulSetName, exists := w.mappings[serviceName]; exists {
+			if scale, err := w.clientset.AppsV1().StatefulSets(service.Namespace).GetScale(ctx, statefulSetName, meta.GetOptions{}); err == nil {
+				replicas := scale.Status.Replicas
+				logrus.WithFields(logrus.Fields{
+					"service":     serviceName,
+					"statefulSet": statefulSetName,
+					"replicas":    replicas,
+				}).Debug("StatefulSet of Service Replicas")
+				if replicas == 0 {
+					if _, err := w.clientset.AppsV1().StatefulSets(service.Namespace).UpdateScale(ctx, statefulSetName, &autoscaling.Scale{
```
itzg, line 232: I'm thinking the reason [this is happening](https://github.com/itzg/mc-router/issues/21#issuecomment-974713622) is because updating the scale is an asynchronous operation and the backend connection is attempted immediately in `findAndConnectBackend`. 

Over in `findAndConnectBackend`, it probably needs to have a retry loop with a delay between each attempt to connect to the backend. It's possibly overkill, but I have been pleased with [this library](https://pkg.go.dev/github.com/cenkalti/backoff/v4) for robust handling of retry logic.

itzg (commented):
This is looking great so far.

docs/k8s-example.yaml
```diff
@@ -9,9 +9,10 @@ kind: ClusterRole
 metadata:
   name: services-watcher
 rules:
-- apiGroups: [""]
-  resources: ["services"]
-  verbs: ["watch","list"]
+- apiGroups: ["", "apps"]
+  resources: ["services", "statefulsets", "statefulsets/scale"]
+  verbs: ["watch","list","get","update"]
+  # how to only watch/list services+statefulsets, and only get+update scale?
```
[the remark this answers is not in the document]
vorburger: FYI when I last worked on this 17 days ago I somehow could not get two separate ClusterRoles working, but agree it is nicer and I will try to rework this again (after PRs 28 & 30), it it was probably just some mix-up. CLI option sure, although why would anyone not want this?! 😈

server/k8s.go
vorburger: @itzg would you be open to leave [resolving this](https://github.com/itzg/mc-router/issues/21#issuecomment-974713622) to a future follow-up PR instead of this one, and consider accepting this once the other feedback here will have been addressed? -- As for the retry logic, I'm thinking that an [exponential backoff](https://en.wikipedia.org/wiki/Exponential_backoff) will delay connecting the user unnecessarily... that kind of approach makes sense to avoid network congestion and backend overload and cascading failures - but that's not our scenario here, do you agree? We just want the (one and only, ever, single) `mc-router` "client" to the connect to that Minecraft server ASAP - which a backoff just delays. Would a basic loop trying for N times with M wait time, perhaps with M=100ms and N globally configurable on the CLI perhaps be both simpler and better?

docs/k8s-example.yaml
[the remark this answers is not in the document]
itzg: Now that I think about it, the role declaration here is only an example so no need to break that out. 

However, I still want the command line option to enable this PR's feature. Also an update to the README stating which permissions are needed for the new feature.

server/k8s.go
itzg: Sure that aspect can be delayed. My suggestion of a retry mechanism didn't mean to imply the use of exponential backoff. Fixed retry interval, as you mentioned, was what I expected.

vorburger:
I've finally rebased this (which was slightly complicated by the "overlapping" great work done by @CDFN in #31).

I hope to work on remaining review feedback (plus some docs about this in the README) within the next week or so.

server/routes.go
itzg: My comment still needs to be addressed.

itzg (commented):
Still waiting on two changes important to me:
- https://github.com/itzg/mc-router/pull/29#discussion_r772746230
- command line option to enable the auto-scaling feature where default is off

server/routes.go
vorburger: Done. (The disadvantage of doing it in `findAndConnectBackend()` instead of here in `FindBackendForServerAddress` is that we can not (easily, as-is) fall back to the default route anymore if wake up fails.)

server/k8s.go
vorburger: Done now. (BTW Note that these statements moved around a bit, due to the conditional flag.)

docs/k8s-example.yaml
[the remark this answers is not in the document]
vorburger: Done. I've reverted the changes to this example YAML (and proposed to remove one of them in #35), and instead added related doc to the README.

vorburger:
> Still waiting on two changes important to me:

Understood, that's why it was marked as Drafted... 🤣 

Both are now done, and I'm now un-drafting this PR to signal that it's now ready for review & merge if OK for you, from my side.

edewit:
:+1:

itzg (commented):
After confirming/adjusting for the `nil` waker, this looks ready to merge.

docs/k8s-deployment.yaml
```diff
@@ -20,7 +20,7 @@ spec:
       containers:
       - image: itzg/mc-router:latest
         name: mc-router
-        args: ["--api-binding", ":8080", "--in-kube-cluster"]
+        args: ["-api-binding", ":8080", "-in-kube-cluster", "-auto-scale-up"]
```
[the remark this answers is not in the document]
vorburger: 👍🏽  (done)

server/routes.go
```diff
@@ -87,20 +88,20 @@ func routesSetDefault(writer http.ResponseWriter, request *http.Request) {
 type IRoutes interface {
 	RegisterAll(mappings map[string]string)
 	// FindBackendForServerAddress returns the host:port for the external server address, if registered.
-	// Otherwise, an empty string is returned
-	// Also returns the normalized version of the given serverAddress
-	FindBackendForServerAddress(serverAddress string) (string, string)
+	// Otherwise, an empty string is returned. Also returns the normalized version of the given serverAddress.
+	// The 3rd value returned is an (optional) "waker" function which a caller must invoke to wake up serverAddress.
```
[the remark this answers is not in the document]
vorburger: ACK, done.

server/routes.go
```diff
@@ -136,11 +145,11 @@ func (r *routesImpl) FindBackendForServerAddress(serverAddress string) (string,
...
 
 	if r.mappings != nil {
-		if route, exists := r.mappings[address]; exists {
-			return route, address
+		if mapping, exists := r.mappings[address]; exists {
+			return mapping.backend, address, mapping.waker
 		}
 	}
-	return r.defaultRoute, address
+	return r.defaultRoute, address, nil
```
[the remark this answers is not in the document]
vorburger: 👍🏽 (done)

server/k8s.go
```diff
@@ -179,6 +238,45 @@ func buildDetails(service *v1.Service, externalServiceName string) *routableServ
...
+						}).Info("StatefulSet Replicas Autoscaled from 0 to 1 (wake up)")
+					} else {
+						return errors.Wrap(err, "UpdateScale for Replicas=1 failed for StatefulSet: "+statefulSetName)
+					}
+				}
+			} else {
+				return errors.Wrap(err, "GetScale failed for StatefulSet: "+statefulSetName)
+			}
+		}
+		return nil
```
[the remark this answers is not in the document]
vorburger: 👍🏽 (done)

README.md
```diff
@@ -157,8 +159,28 @@ kubectl apply -f https://raw.githubusercontent.com/itzg/mc-router/master/docs/k8
...
+
+```yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: services-watcher
+rules:
+- apiGroups: ["", "apps"]
+  resources: ["services", "statefulsets", "statefulsets/scale"]
+  verbs: ["watch","list","get","update"]
```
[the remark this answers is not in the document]
vorburger: 👍🏽 (done)

vorburger:
> After confirming/adjusting for the `nil` waker, this looks ready to merge.

Done! (Do have a look at if I understood and fixed it correctly.)

itzg (approved):

PR #33, opened 2021-12-07
misc: Disable organizing imports in Visual Studio Code (see #28)

vorburger:
see our discussion in #28

This is actually NOT my recommended / favourite solution to that problem, but let's see your response in #28.

Alternatively, if you don't want to change import ordering on your side, then do merge this, at least for now, as it makes it much easier for me.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,10 @@
+{
+    // https://github.com/itzg/mc-router/pull/28
+    "go.formatTool": "gofmt",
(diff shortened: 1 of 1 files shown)


itzg:
Given option b in #28, then this PR isn't needed, right?

vorburger:
> Given option b in #28, then this PR isn't needed, right?

Yup, I'll close this and update #28 with a full reformat of all imports.

PR #35, opened 2021-12-28
Remove un-used k8s-example.yaml

vorburger:
This is virtually identical to the k8s-example-auto.yaml, and that one, contrary to this one, IS referenced in the README.

I suspect that this may be a left-over example from before auto-discovery (but now both of these examples illustrate that, so this one is now moot).


vorburger:
I've run into this while writing the doc for #29. (I'll remove cahnges to this file from that PR, assuming it's going to be removed.)

PR #97, opened 2022-07-12
Trim root zone indicator from server address

itzg:
Trims the root zone indicator from an incoming server address, such as the trailing dot from `mc.example.com.`. [See this](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) for more info about FQDNs and the root zone. 

cc @Merith-TK


itzg:
FYI, null delimiters in the server address were already handled

https://github.com/itzg/mc-router/blob/1a873cb58a2c838c4c54cbabfc6dcee27f384be4/server/routes.go#L171

PR #98, opened 2022-07-15
depends_on for router

mhkarimi1383:
router should be available when main servers are okay :)

### DIFF
--- a/docs/docker-compose.yml
+++ b/docs/docker-compose.yml
@@ -12,7 +12,7 @@ services:
       TYPE: FORGE
   router:
     image: ${MC_ROUTER_IMAGE:-itzg/mc-router}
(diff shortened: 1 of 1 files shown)


docs/docker-compose.yml
```diff
@@ -12,6 +12,9 @@ services:
       TYPE: FORGE
   router:
     image: ${MC_ROUTER_IMAGE:-itzg/mc-router}
+    depens_on:
```
itzg, line 15:
```suggestion
    depends_on:
```

itzg (commented):
Thanks, that's a helpful change for the example. Just a typo fix and I'll get this merged.

itzg (approved):

PR #103, opened 2022-07-20
Add support for docker swarm service discovery

### LINKED ISSUE #60: Docker Swarm Support planend
To explain, I am running a docker swarm cluster, using the mc routuer works as intended, but I have a suggestion. Could you add dynamic support like with kubernetes but with docker labels (like traefik does ist). so you set a label called "mc-router=enabled" and "mc-router=example.domain@example.com"  and now its routs the traffic of the container where the labels are set.

n1xx1:
Similar to the K8s support for service discovery, inspired by how traefik does it. Should resolve #60.

I've added a `--in-docker-swarm` flag to enable it, you have to deploy the container binding `/var/run/docker.sock` of the host to `/var/run/docker.sock`, and add these flags to the service: `mc-router.host` with the list of domains to expose, `mc-router.port` in case your server isn't deployed to 25565.

WIP because it's untested. Let me know what you think.


server/docker.go
```diff
@@ -0,0 +1,203 @@
...
+	// The key in mappings is a Service, and the value the StatefulSet name
+	mappings map[string]string
+
+	client        *client.Client
+	contextCancel context.CancelFunc
+}
+
+const (
+	// TODO: maybe expose
+	dockerConfigHost = "unix:///var/run/docker.sock"
```
itzg, line 38: This one seems fine to hardcode

itzg (commented):
Cool. That's looking really good.

n1xx1:
I updated the PR, I'm missing some automated tests, but I'm currently using it in my server without any problems.

Meanwhile can you approve the workflows? Just to see if there are any issues with the go.mod changes. 

Thanks!

itzg (commented):
I fixed up the go.sum, so now the build succeeds.

Add a brief description in the README section you started and then this looks ready to merge.

itzg (approved):
Awesome. Ready for merge? BTW, this and kube support are difficult to unit test, so I'm good with your manual testing effort.

n1xx1:
I did a run of manual testing on my server and found out I forgot a `ok = true`. After this change it should be all good and I also implemented a default server label.

itzg:
Sorry for the delay in merging this. It is now released in https://github.com/itzg/mc-router/releases/tag/1.16.0

Issue #144, opened 2022-12-01
Question: Persistent routes


LasseR15:
Hi,

I'm tinkering a bit with the REST API right now and really like this project.
Is it possible to persist the routes added or changed via the API so that they are preserved even after a restart?

itzg:
Not yet but that would be a great enhancement. 

If you're interested in doing that, I would suggest persisting to a JSON file that is named from a user argument. That way the feature can serve double duty as a way to pre-load the routes from a file.

LasseR15:
Do you mean an argument at startup like the argument for API bindings (`-api-binding`)? So for example `-routes-config` (env: ROUTES_CONFIG) for the path to the configuration file (or just the name)?

itzg:
Yes, that's what I was thinking. And a path seems more versatile. That way it can also be just a filename relative to working directory.

LasseR15:
Alright, I'll look into it this weekend and open a pull request as soon as I'm ready.

itzg:
Implemented by #145

PR #145, opened 2022-12-04
Persistent routes from REST API

LasseR15:
Issue: #144 
Added functionality to persist the routes from the API in a json formatted config file

### DIFF
--- a/server/routes_config.go
+++ b/server/routes_config.go
@@ -3,28 +3,135 @@ package server
 import (
 	"encoding/json"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	"io/fs"
 	"os"
+	"sync"
 )
 
+type IRoutesConfig interface {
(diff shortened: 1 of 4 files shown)


server/routes_config.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+func (r *routesConfigImpl) readRoutesConfigFile() (routesConfigStructure, error) {
+	r.RLock()
+	defer r.RUnlock()
+
+	config := routesConfigStructure{
+		"",
+		make(map[string]string),
+	}
```
itzg, line 143: I don't believe it is necessary to prep the empty content. I think just a `var config routesConfigStucture` would suffice.

itzg (approved):
Thanks for a great feature contribution! 

My question in there is just a curiosity/stylistic thing.

itzg:
This is now released in https://github.com/itzg/mc-router/releases/tag/1.18.0

PR #195, opened 2023-05-24
Possible solution for wildcard fixing

rainbowdashlabs:
First of all: I never used GO before. So see this more as a proposal than a proper implementation. Feel free to make changes.

Background of this is that services like tcp shield add more context to the domain. If I route a domain like mc.example.com I get something like 
```
mc.example.com///127.0.0.1:25565///1684918045///MGQCMBSW001TpRFxZBMq6nA/OS8y1B8S3sUc9CpAeFb7lzdvtvP+vqjjI8mMD1sQ1njiFQIwOyqxfe6HpxT+GYkOJB5n0VhDruEWMQySf8GcudKmTArMvPEEtAtQWUbO8hkSx++y
```
This is an ever changing value which has some meaning for tcp shield. So far it seems to be unique for every single connection.

At the current state tcp shield cant be used in front of mc-router since those urls are impossible to match.
Allowing to do wild card matches like `mc.example.com/* Would allow the router to accept those connections as well. 

Also the docker build fails for me, but it doesnt look like an compiler error, so I am quite unsure how to resolve it, but I also noticed that the test build also failed the last times, so it is probably not related to my changes? Therefore this change is still untested.

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -12,6 +12,8 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+var tcpShieldPattern = regexp.MustCompile("///.*")
+
 func init() {
 	apiRoutes.Path("/routes").Methods("GET").
 		Headers("Accept", "application/json").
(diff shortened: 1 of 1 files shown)


itzg:
Thanks for addressing this. Since TCP shield seems to have a nice, clean syntax to the host enrichment, I'm thinking we could just have it always strip away "///.*" from any inbound host request. As it is, the change won't help users with dynamic routes, which I'm guessing much is much more prevalent.

Yes, sorry for the latest build issues. This has unfortunately fallen low on my long list of things to do. I'll try to address those shortly. I know you're new to Go, but will also be worthwhile to install Go locally (which is very easy) and can build/test with 

```
go test ./...
```

rainbowdashlabs:
The tests are passing. I adjusted the code to simply strip the suffix in any request.

itzg (approved):
Thanks for the rapid updates.

PR #270, opened 2024-02-19
auto scale up and down with many named backends in k8s

aapjeisbaas:
I created a cron job to scale down empty servers.
There is room for improvement but it runs stable and that's the most important thing I guess.

This YAML is an example and is not intended to be directly applied.
It consists of 3 parts
1. the mc-router with service account and service 
2. the shutdown cronjob with service account
3. the actual minecraft server with service and storage

Part 3 is the only part you need to replicate for the number of servers you want.

### DIFF
--- a/docs/k8s-autoscale.yaml
+++ b/docs/k8s-autoscale.yaml
@@ -1,0 +1,278 @@
+
+# This YAML is an example and is not intended to be directly applied.
+# It consists of 3 parts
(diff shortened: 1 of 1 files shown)


itzg (approved):
Very nice! Thanks.

itzg:
@aapjeisbaas can you provide some advice on #378?

PR #316, opened 2024-08-04
Add Service Discovery option for non-Swarm Docker

### LINKED ISSUE #110: Enable docker (non swarm) auto discovery
Docker swarm support is added (#60) and it should work with non-swarm deployments as well, however if you start it with --in-docker-swarm you get this:

time="2022-08-15T17:23:00Z" level=fatal msg="Unable to start docker swarm integration" error="Error response from daemon: This node is not a swarm manager. Use \"docker swarm init\" or \"docker swarm join\" to connect this node to swarm and try again."

I suggest adding --in-docker to enable autodiscovery in a plain docker engine deployment.

devmattrick:
This PR adds a basic implementation of service discovery for non-Swarm Docker deployments. You can see an example of this in action in [`sd-docker.docker-compose.yml`](https://github.com/itzg/mc-router/compare/master...devmattrick:devmattrick/docker-service-discovery?expand=1#diff-0f51bf91591ee773dcb56dc5da6a71a8a337d61c335632da62ba1997d79f386c).

This should fix #110. Feel free to send any questions or feedback my way!

### DIFF
--- a/server/docker.go
+++ b/server/docker.go
@@ -15,12 +15,11 @@ import (
 )
 
 type IDockerWatcher interface {
-	Start(timeoutSeconds int, refreshIntervalSeconds int) error
+	Start(socket string, timeoutSeconds int, refreshIntervalSeconds int) error
 	Stop()
 }
 
 const (
-	DockerConfigHost         = "unix:///var/run/docker.sock"
 	DockerAPIVersion         = "1.24"
 	DockerRouterLabelHost    = "mc-router.host"
 	DockerRouterLabelPort    = "mc-router.port"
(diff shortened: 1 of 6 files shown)


itzg (commented):
This is great. 

Can you also update the README? Feel free to repurpose the Docker Swarm section and then adding back a Swarm section that says it's all the same except labels on services and the config flag to enable.

server/docker.go
```diff
@@ -3,34 +3,22 @@ package server
...
-var DockerWatcher IDockerWatcher = &dockerWatcherImpl{}
-
-type dockerWatcherImpl struct {
-	sync.RWMutex
-	client        *client.Client
-	contextCancel context.CancelFunc
-}
-
 const (
 	DockerConfigHost         = "unix:///var/run/docker.sock"
```
itzg, line 23: It would be nice for this to be configurable, such as for local testing on Windows...but I don't feel strongly about.

devmattrick: Done!

devmattrick:
Went ahead and updated the README and added an option to configure the Docker Socket. Let me know if there's any changes to the README you'd like.

itzg (approved):
Thanks for the quick follow-up.

Issue #311, opened 2024-07-16
mc-router Fails to Route Requests and Auto-Scale Minecraft Server in kubernetes


csabca83:
# Description
## Describe the bug

I am facing an issue with mc-router where requests routed through the mc-router service fail to reach the backend Minecraft server. Additionally, the auto-scaling functionality does not seem to be triggered, and clients trying to connect to the Minecraft server time out.

# To Reproduce

## Steps to reproduce the behavior:

Deploy mc-router and Minecraft server using the provided Kubernetes manifests.
Expose mc-router on a NodePort for HTTP (30013) and TCP (30012) for Minecraft.
Attempt to connect to the Minecraft server via the mc-router service on port 30013.
## Expected behavior

The mc-router should route the Minecraft traffic to the backend Minecraft server, and auto-scaling should be triggered if necessary.

# Environment:

Kubernetes Version: [v1.30.2]
mc-router Version: 1.21.0
mcserver Version: 2024.7.0
# Logs:

Whenever I try to reach out through 30013 there are no logs, I can only see a timeout exception on the client side.
However if I try to connect directly through 30012 which is the direct port to the mcserver via the mc-router it works if the mcserver scale is at 1. However if the scale is 0 it won't scale it, just return an error saying that the backend server isn't reachable. Either one would be fine for me as long as 0->1 scaling would happen.
```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: mc-router
  namespace: mcserver
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: mc-router-services-watcher-crole
rules:
- apiGroups: [""]
  resources: ["services"]
  verbs: ["watch", "list"]
- apiGroups: ["apps"]
  resources: ["statefulsets", "statefulsets/scale"]
  verbs: ["watch", "list"]
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: mc-router-services-watcher
  namespace: mcserver
subjects:
- kind: ServiceAccount
  name: mc-router
  namespace: mcserver
roleRef:
  kind: ClusterRole
  name: mc-router-services-watcher-crole
  apiGroup: rbac.authorization.k8s.io
---
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  name: mc-router-stsscaler-role
  namespace: mcserver
rules:
- apiGroups: ["apps"]
[...]

PR #317, opened 2024-08-06
Update documentation - autoscaling requirements

csabca83:
This PR updates the documentation to raise awareness of the requirements for StatefulSet autoscaling in Kubernetes to function properly.

#311

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,8 +245,7 @@ rules:
 ```
 
 Make sure to set `StatefulSet.metadata.name` and `StatefulSet.spec.serviceName` to the same value;
-otherwise, autoscaling will not trigger (`mc-router.itzg.me/externalServerName` annotation is also required for it to work,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -244,6 +244,29 @@ rules:
   verbs: ["watch","list","get","update"]
 ```
 
+Make sure to set `StatefulSet.metadata.name` and `StatefulSet.spec.serviceName` to the same value;
+otherwise, autoscaling will not trigger (`mc-router.itzg.me/externalServerName` annotation is also required for it to work,
+however if you are not using a DNS you can specify an IP address as well `"mc-router.itzg.me/externalServerName": "1.2.3.4"`):
```
itzg, line 249: Please remove the part that's in parentheses. It's already documented above in the kubernetes section. You can add the info about using an IP address there.

itzg (commented):
Thanks

README.md
csabca83: Makes sense, just removed it now :)

itzg (approved):

PR #383, opened 2025-04-15
Add support for ghcr.io image tags

### LINKED ISSUE #379: Request: Publish images to alternative registry
Currently, mc-router is only published to docker. With new rate-limiting restrictions coming down, would it be possible to publish to additional registries, since you can package and push to multiple registries at the same time?

Typically, I prefer ghcr, since we're already on github, but something like quay would also be great.

You already do this in the https://github.com/itzg/docker-minecraft-server repo.

kashalls:
Fixes #379.

Have not fully tested docker hub deployment.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,7 +10,7 @@ jobs:
   release:
     uses: itzg/github-workflows/.github/workflows/go-with-releaser-image.yml@main
     with:
-      go-version: "1.24.1"
+      go-version: "1.24.2"
     secrets:
       image-registry-username: ${{ secrets.DOCKERHUB_USERNAME }}
       image-registry-password: ${{ secrets.DOCKERHUB_TOKEN }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,4 +11,4 @@ jobs:
   build:
     uses: itzg/github-workflows/.github/workflows/go-test.yml@main
     with:
-      go-version: "1.24.1"
+      go-version: "1.24.2"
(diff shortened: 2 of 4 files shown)


itzg (commented):
Please put the filename back to what it was. It made the diff unusable.

kashalls:
> Also, I’m a little disappointed that the copy-paste approach was used when I said in the issue that I didn’t want to go that way. Instead I wanted to use goreleaser’s ko support.

I'm sorry to disappoint you on this front, I don't like to use goreleaser as it follows a template. It is like a glue that could be deprecated in the future. I prefer to have full control/auditability of the pipeline. As it is right now, testing the goreleaser threw some warnings about your config being out of date already.

> Is your need for GHCR so urgent that it warrants this expedited approach? What’s wrong with using the Docker Hub published images? Is it just because of the “rate limit” they changed?

I hit the docker hub ratelimit almost every time myself between running Renovate & caching images for internal cluster use.

itzg:
Wow, only 16 (100 per 6 hours) pulls per hour unauthenticated:

https://docs.docker.com/docker-hub/usage/

I didn't realize they tightened down that far. I'll merge this to get GHCR going. 

Later I'll explore using ko outside of goreleaser to minimize goreleaser involvement. For me, goreleaser still solves the tediousness of
- multi-arch builds
- creating archive of each
- uploading to a github release
- pushing one or more of those to a scoop bucket repo
- etc

I don't want to bother writing a script for all that.

itzg:
Oh, I forgot to check if a docker login had been added for GHCR

itzg:
Image published: https://github.com/itzg/mc-router/pkgs/container/mc-router

PR #388, opened 2025-04-18
Do not autoscale up on status requests

### LINKED ISSUE #387: Auto scale up only on player connect
Currently with `--auto-scale-up` enabled, a server will be scaled [as soon as a backend connection is attempted](https://github.com/itzg/mc-router/blob/84cbe38ef3b2a68acdc97433d3407d90bfed0d64/server/connector.go#L262). However, this causes servers to be unnecessarily scaled up if a user has them saved in their server list because the Minecraft client attempts to ping all servers in the list. Below is a screenshot of two servers I have behind the `mc-router` deployed on Kubernetes. The two server StatefulSets were both at 0 replicas before opening Minecraft, and by simply opening this server screen both were started up:

Image

It would be great to introduce an option that only scales up a server if a player is actively attempting to join or playing on the server. Otherwise servers may be scaled up even if a player was intending to play on a completely different server. I understand if this may not be easily feasible but would significantly cut down on unnecessary server starts. Thanks so much!

Sammcb:
Fixes #387 

Looking at the Minecraft protocol documentation, I saw the [handshake request sends info about the state of following requests](https://minecraft.wiki/w/Java_Edition_protocol#Handshake). I saw this was already being parsed so I simply passed that through to the `findAndConnectToBackend` and only scale if that value is not a `status` request (for the legacy server ping, I hard-code the `NextState` to match the `status` state of the current protocol).

I tested this by building the image and loading it into a local Kubernetes cluster with two Minecraft servers behind the `mc-router`. I then launched Minecraft and confirmed that simply opening the server list did not scale up either server. I attempted to connect to `Server A` and saw the StatefulSet scale up. I then attempted to connect to `Server B` and confirmed that StatefulSet also scaled up.

I'm not super familiar with the Minecraft protocol and am unsure if there would be any negative side effects of this change. If so, I'm happy to put this behind another config flag!

### DIFF
--- a/server/connector.go
+++ b/server/connector.go
@@ -223,8 +223,9 @@ func (c *Connector) HandleConnection(ctx context.Context, frontendConn net.Conn)
 			Debug("Got handshake")
 
 		serverAddress := handshake.ServerAddress
+		nextState := handshake.NextState
 
-		c.findAndConnectBackend(ctx, frontendConn, clientAddr, inspectionBuffer, serverAddress)
+		c.findAndConnectBackend(ctx, frontendConn, clientAddr, inspectionBuffer, serverAddress, nextState)
 	} else if packet.PacketID == mcproto.PacketIdLegacyServerListPing {
 		handshake, ok := packet.Data.(*mcproto.LegacyServerListPing)
 		if !ok {
@@ -242,8 +243,9 @@ func (c *Connector) HandleConnection(ctx context.Context, frontendConn net.Conn)
 			Debug("Got legacy server list ping")
 
 		serverAddress := handshake.ServerAddress
+		nextStateStatus := 1
 
-		c.findAndConnectBackend(ctx, frontendConn, clientAddr, inspectionBuffer, serverAddress)
+		c.findAndConnectBackend(ctx, frontendConn, clientAddr, inspectionBuffer, serverAddress, nextStateStatus)
 	} else {
 		logrus.
 			WithField("client", clientAddr).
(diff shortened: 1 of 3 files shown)


Sammcb:
@itzg Would love a review on this. Open to any suggestions or feedback!

server/connector.go
```diff
@@ -255,10 +257,11 @@ func (c *Connector) HandleConnection(ctx context.Context, frontendConn net.Conn)
 }
 
 func (c *Connector) findAndConnectBackend(ctx context.Context, frontendConn net.Conn,
-	clientAddr net.Addr, preReadContent io.Reader, serverAddress string) {
+	clientAddr net.Addr, preReadContent io.Reader, serverAddress string, nextState int) {
 
 	backendHostPort, resolvedHost, waker := Routes.FindBackendForServerAddress(ctx, serverAddress)
-	if waker != nil {
+	nextStateStatus := 1
```
itzg, line 263: Please declare a const of type State over in https://github.com/itzg/mc-router/blob/master/mcproto/types.go that is called StateStatus and then use that here and in the SLP code above. As it currently is, it doesn't look like a protocol defined value.

Sammcb: Thanks for pointing that out, that's much clearer.

itzg (approved):
Thanks for the update. Agreed that looks a lot better. 

So, the only downside of this change (when using auto scale and scaled down) is that the server will appear non-responsive in the server list?

itzg:
Released in https://github.com/itzg/mc-router/releases/tag/1.26.2

PR #391, opened 2025-04-20
Add option to emit metrics and logs when players connect to the router

### LINKED ISSUE #390: Log and emit metrics on player login
While the current behavior of `--auto-scale-up` is useful, it is not able to check against a server allowlist to confirm a player is allowed to join before scaling up the server. Emitting metrics/logs on which players are attempting to connect to specific backend servers would allow users define custom scalers (using a tool such as KEDA) for scaling up backend servers only when allowed players are connecting.

This telemetry could also be useful for monitoring setups, etc.

Sammcb:
Fixes #390 

Note, `discard` and `expvar` backends do not "officially" support the emitted metrics. `discard` for obvious reasons, but `expvar` because the metrics do not support label values.


Sammcb:
@itzg Open to any suggestions/feedback for docs/testing/naming/structure, etc!

itzg:
Sorry, you'll need to adjust some amount of this after my upcoming PR. I don't get much energy to clean up the mc-router code, so I did a bit more than usual.

Sammcb:
No worries, happy to rebase once your PR is submitted!

itzg:
I have merged my PR that conflicts a bit with yours. It's probably best to remove the `mcproto` changes from yours and adjust the other changes around that.

Sammcb:
@itzg I believe this is ready for review now. Please let me know if you have any preference for the metric names, etc. I was able to test this with the `prometheus` backend but not the `influxdb` backend as I don't have an instance to write to.

Please let me know if there's any additional testing you would like me to try to do!

mcproto/read.go
```diff
@@ -45,7 +45,8 @@ func ReadPacket(reader *bufio.Reader, addr net.Addr, state State) (*Packet, erro
 		return nil, err
 	}
 
-	packet := &Packet{Length: frame.Length}
+	// Packet length is frame length (bytes for packetID and data) plus bytes used to store the frame length data
+	packet := &Packet{Length: frame.Length + PacketLengthFieldBytes}
```
itzg, line 49: Yeah, I was just noticing that in the protocol docs. Does it actually make a difference later in the code from here?

itzg (approved):
Thanks! I liked all the metric names. Ready for me to merge this?

Sammcb:
Yep!

mcproto/read.go
Sammcb: I checked and there weren't any usages of it. Just something I had noticed when making the initial PR. Personally I think it's a little clearer as this is the actual packet size but happy to change it back if you would prefer.

itzg:
This is now released in 1.27.1

PR #392, opened 2025-04-20
Add connection webhook notifications

itzg:


server/webhook_notifier.go
```diff
@@ -0,0 +1,157 @@
...
+		ClientPort:    clientAddr.(*net.TCPAddr).Port,
+		ServerAddress: serverAddress,
+		PlayerInfo:    playerInfo,
+		Error:         "No backend found",
+	}
+
+	return w.send(ctx, payload)
+}
+
+func (w *WebhookNotifier) NotifyFailedBackendConnection(ctx context.Context, clientAddr net.Addr, serverAddress string,
```
kashalls, line 83: There has to be a better way to do this. You shouldn't have functions that return different responses. The best case should have one function and when its called it should provide only the information it has at the time.

You're kind of repeating yourself here.

server/webhook_notifier.go
```diff
@@ -0,0 +1,157 @@
...
+const (
+	WebhookStatusMissingBackend          = "missing-backend"
+	WebhookStatusFailedBackendConnection = "failed-backend-connection"
+	WebhookStatusSuccess                 = "success"
+)
+
+type WebhookNotifierPayload struct {
+	Event           string      `json:"event"`
+	Status          string      `json:"status"`
+	ClientHost      string      `json:"client-host"`
```
kashalls, line 46: I would include client host and port under player info as mentioned earlier. If player info like name or uuid is unavailable, I would just set it to nullish values (empty string: "", uuid of 0's)

server/connector.go
```diff
@@ -222,10 +242,26 @@ func (c *Connector) HandleConnection(ctx context.Context, frontendConn net.Conn)
...
+				return
+			}
+			logrus.
+				WithField("client", clientAddr).
+				WithField("userInfo", userInfo).
+				Debug("Got user info")
+
+			c.findAndConnectBackend(ctx, frontendConn, clientAddr, inspectionBuffer, handshake.ServerAddress, userInfo, handshake.NextState)
+		} else {
+			c.findAndConnectBackend(ctx, frontendConn, clientAddr, inspectionBuffer, handshake.ServerAddress, nil, handshake.NextState)
```
kashalls, line 262: You could just pass an empty user information object here and avoid calling the same function twice here.

kashalls (commented):
Apologies for the wall of comments, I think these should be looked into further. Some of the code produces unwanted recursions or (DRY).

server/connector.go
itzg: It needs to be nil for the JSON field omission to work as I want. With that said, I can shift it up to a var to avoid the ugly double-call.

server/webhook_notifier.go
itzg: The originating client IP+port is always available even during a server list/ping operation and refers to the client machine. Player info is separately provided to allow for webhook consumer to differentiate a ping from a user attempting a session.

server/webhook_notifier.go
itzg: That's a bit rude.  I thought API was a balance of cleanly documented yet filling out respective parts of the payload.

kashalls: > That's a bit rude. 

I apologize, I did not mean to have it sound like that. Working with Kubernetes + YAML, the goal was always to avoid repeating yourself wherever possible. 

The best approach would be to create a object that is effectively "empty" and fill in information as you have it. When it's sent to the audit, if it's not filled out the server did not have it.

server/webhook_notifier.go
kashalls: > The originating client IP+port is always available even during a server list/ping operation and refers to the client machine. Player info is separately provided to allow for webhook consumer to differentiate a ping from a user attempting a session.

If that is the case then, it would probably be better to include under a "client" object. I believe Javascript wouldn't be able to reference this in object notation and would instead have to handle it as `const host = body["client-host"]`

image

cmd/mc-router/metrics.go
```diff
@@ -24,14 +24,26 @@ type MetricsBuilder interface {
...
 		return &expvarMetricsBuilder{}
-	case "prometheus":
+	case MetricsBackendPrometheus:
 		return &prometheusMetricsBuilder{}
-	case "influxdb":
+	case MetricsBackendInfluxDB:
 		return &influxMetricsBuilder{config: config}
+	case MetricsBackendDiscard:
+		return &discardMetricsBuilder{}
 	default:
```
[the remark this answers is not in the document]
itzg: Yes, I investigated this and I have ended up with too many call sites. The compromise was to leave the no-op, discard implementation in place eliminate resource usage when no metrics backend is used.

README.md
```diff
@@ -353,6 +357,69 @@ From those logs, locate the `ngrokUrl` parameter from the "Listening" info log m
...
+
+Refer to [usage](#usage) for `--webhook-*` argument descriptions.
+
+### Sample payloads
+
+```json
+{
+  "event": "connecting",
+  "status": "success",
+  "client-host": "127.0.0.1",
```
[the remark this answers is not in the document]
itzg: I'll compromise and put it under a `client` object.

README.md
```diff
@@ -74,6 +74,10 @@ Routes Minecraft client connections to backend servers based upon the requested
     	Send PROXY protocol to backend servers (env USE_PROXY_PROTOCOL)
   -version
     	Output version and exit (env VERSION)
+  -webhook-require-user
+    	If set, the webhook will only be called if the user is connecting rather than just server list/ping (env WEBHOOK_REQUIRE_USER)
+  -webhook-url string
+    	If set, a webhook will be POST'ed to the given endpoint for connection status notifications (env WEBHOOK_URL)
```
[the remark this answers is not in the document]
itzg: I like that

server/webhook_notifier.go
itzg: Moved into a `client` object.

server/webhook_notifier.go
itzg: Yeah, I've been developing code for 30 years...I know about DRY 😄I'll probably collapse the two failure modes into one method, but my coding preference won't allow much more than that.

itzg: Since I have designed the `ConnectionNotifier` interface to be generic, combining those methods would instead require the caller to pass in a "status string". Maybe there's a Discord notifier and it wants to use an emoji instead of a particular string to convent a missing backend.

When you've written code as long as I have you start to find the balances between repeating a few lines of code and the maintainability and extensibility of the code around it.

PR #396, opened 2025-04-23
Call webhook on player disconnect

UrbaneChimp:


server/webhook_notifier.go
```diff
@@ -109,6 +110,24 @@ func (w *WebhookNotifier) NotifyConnected(ctx context.Context, clientAddr net.Ad
 	return w.send(ctx, payload)
 }
 
+func (w *WebhookNotifier) NotifyDisconnected(ctx context.Context, clientAddr net.Addr, serverAddress string, playerInfo *PlayerInfo, backendHostPort string) error {
+	if w.requireUser && playerInfo == nil {
+		return nil
+	}
+
+	payload := &WebhookNotifierPayload{
+		Event:           WebhookEventConnecting,
```
itzg, line 119: Actually, let's go with an event string of "disconnect", so declare a `WebhookEventDisconnecting` to pair up with the other.

server/webhook_notifier.go
```diff
@@ -29,6 +29,7 @@ const (
 	WebhookStatusMissingBackend          = "missing-backend"
 	WebhookStatusFailedBackendConnection = "failed-backend-connection"
 	WebhookStatusSuccess                 = "success"
+	WebhookStatusDisconnected            = "disconnected"
```
itzg, line 32: ...and rather a new status value

server/webhook_notifier.go
```diff
@@ -109,6 +110,24 @@ func (w *WebhookNotifier) NotifyConnected(ctx context.Context, clientAddr net.Ad
...
 
+func (w *WebhookNotifier) NotifyDisconnected(ctx context.Context, clientAddr net.Addr, serverAddress string, playerInfo *PlayerInfo, backendHostPort string) error {
+	if w.requireUser && playerInfo == nil {
+		return nil
+	}
+
+	payload := &WebhookNotifierPayload{
+		Event:           WebhookEventConnecting,
+		Timestamp:       time.Now(),
+		Status:          WebhookStatusDisconnected,
```
itzg, line 121: ...use WebhookStatusSuccess here. Maybe we could use a non-success status if the connection was dropped prematurely...however the code probably can't detect that difference right now.

itzg (commented):
Awesome. Thanks for finishing out the feature...and so quickly after the first round was added.

Just some tweaks on the event/status please.

itzg (approved):
Thanks!

itzg:
Now released in 1.27.2

PR #405, opened 2025-04-29
Add auto scale down option

### LINKED ISSUE #384: Auto scale down
I really appreciate the auto scale up feature. I know in the docs there is a [sample CronJob](https://github.com/itzg/mc-router/blob/0868f04de1f3bd273e64bd0a213988727de788f4/docs/k8s-autoscale.yaml#L138) that can be included for scaling servers back down. I was wondering if this functionality could be added natively to the `mc-router`, where it would also set server StatefulSet replicas to 0 after some configurable amount of time has passed without any connections.

Thanks so much for these amazing projects, really appreciate how smooth running Minecraft servers is with these!

Sammcb:
Fixes #384

### DIFF
--- a/cmd/mc-router/main.go
+++ b/cmd/mc-router/main.go
@@ -59,6 +59,8 @@ type Config struct {
 	RoutesConfig          string `usage:"Name or full path to routes config file"`
 	NgrokToken            string `usage:"If set, an ngrok tunnel will be established. It is HIGHLY recommended to pass as an environment variable."`
 	AutoScaleUpAllowDeny  string `usage:"Path to config for server allowlists and denylists. If -auto-scale-up is enabled and a global/server entry is specified, only players allowed to connect to the server will be able to trigger a scale up"`
+	AutoScaleDown         bool   `default:"false" usage:"Decrease Kubernetes StatefulSet Replicas (only) from 1 to 0 on respective backend servers after there are no connections"`
+	AutoScaleDownAfter    string `default:"10m" usage:"Server scale down delay after there are no connections"`
 
 	ClientsToAllow []string `usage:"Zero or more client IP addresses or CIDRs to allow. Takes precedence over deny."`
 	ClientsToDeny  []string `usage:"Zero or more client IP addresses or CIDRs to deny. Ignored if any configured to allow"`
@@ -123,6 +125,14 @@ func main() {
 	defer cancel()
 
 	metricsBuilder := NewMetricsBuilder(config.MetricsBackend, &config.MetricsBackendConfig)
+
+	downScalerEnabled := config.AutoScaleDown && (config.InKubeCluster || config.KubeConfig != "")
+	downScalerDelay, err := time.ParseDuration(config.AutoScaleDownAfter)
+	if err != nil {
+		logrus.WithError(err).Error("Unable to parse down scale after duration")
+	}
+	server.DownScaler = server.NewDownScaler(ctx, downScalerEnabled, downScalerDelay)
+
 
 	c := make(chan os.Signal, 1)
 	signal.Notify(c, syscall.SIGINT, syscall.SIGTERM)
--- a/server/connector.go
+++ b/server/connector.go
@@ -342,7 +342,7 @@ func (c *Connector) readUserInfo(bufferedReader *bufio.Reader, clientAddr net.Ad
 func (c *Connector) findAndConnectBackend(ctx context.Context, frontendConn net.Conn,
 	clientAddr net.Addr, preReadContent io.Reader, serverAddress string, playerInfo *PlayerInfo, nextState mcproto.State) {
 
-	backendHostPort, resolvedHost, waker := Routes.FindBackendForServerAddress(ctx, serverAddress)
+	backendHostPort, resolvedHost, waker, _ := Routes.FindBackendForServerAddress(ctx, serverAddress)
 	if waker != nil && nextState > mcproto.StateStatus {
 		serverAllowsPlayer := c.autoScaleUpAllowDenyConfig.ServerAllowsPlayer(serverAddress, playerInfo)
 		logrus.
@@ -352,6 +352,15 @@ func (c *Connector) findAndConnectBackend(ctx context.Context, frontendConn net.
 			WithField("serverAllowsPlayer", serverAllowsPlayer).
 			Debug("checked if player is allowed to wake up the server")
 		if serverAllowsPlayer {
+			// Cancel down scaler if active before scale up
+			DownScaler.Cancel(serverAddress)
+			// Needs to trigger after defer where c.activeConnections is decremented
+			defer func() {
+				activeConnections := atomic.LoadInt32(&c.activeConnections)
+				if activeConnections <= 0 {
+					DownScaler.Begin(serverAddress)
+				}
+			}()
 			if err := waker(ctx); err != nil {
 				logrus.WithFields(logrus.Fields{"serverAddress": serverAddress}).WithError(err).Error("failed to wake up backend")
 				c.metrics.Errors.With("type", "wakeup_failed").Add(1)
--- a/server/docker.go
+++ b/server/docker.go
@@ -41,6 +41,12 @@ func (w *dockerWatcherImpl) makeWakerFunc(_ *routableContainer) func(ctx context
 	}
 }
 
+func (w *dockerWatcherImpl) makeSleeperFunc(_ *routableContainer) func(ctx context.Context) error {
+	return func(ctx context.Context) error {
+		return nil
+	}
+}
+
 func (w *dockerWatcherImpl) Start(socket string, timeoutSeconds int, refreshIntervalSeconds int) error {
 	var err error
 
@@ -75,7 +81,7 @@ func (w *dockerWatcherImpl) Start(socket string, timeoutSeconds int, refreshInte
 	for _, c := range initialContainers {
 		containerMap[c.externalContainerName] = c
 		if c.externalContainerName != "" {
-			Routes.CreateMapping(c.externalContainerName, c.containerEndpoint, w.makeWakerFunc(c))
+			Routes.CreateMapping(c.externalContainerName, c.containerEndpoint, w.makeWakerFunc(c), w.makeSleeperFunc(c))
 		} else {
 			Routes.SetDefaultRoute(c.containerEndpoint)
 		}
@@ -97,7 +103,7 @@ func (w *dockerWatcherImpl) Start(socket string, timeoutSeconds int, refreshInte
 						containerMap[rs.externalContainerName] = rs
 						logrus.WithField("routableContainer", rs).Debug("ADD")
 						if rs.externalContainerName != "" {
-							Routes.CreateMapping(rs.externalContainerName, rs.containerEndpoint, w.makeWakerFunc(rs))
+							Routes.CreateMapping(rs.externalContainerName, rs.containerEndpoint, w.makeWakerFunc(rs), w.makeSleeperFunc(rs))
 						} else {
 							Routes.SetDefaultRoute(rs.containerEndpoint)
 						}
@@ -105,7 +111,7 @@ func (w *dockerWatcherImpl) Start(socket string, timeoutSeconds int, refreshInte
 						containerMap[rs.externalContainerName] = rs
 						if rs.externalContainerName != "" {
 							Routes.DeleteMapping(rs.externalContainerName)
-							Routes.CreateMapping(rs.externalContainerName, rs.containerEndpoint, w.makeWakerFunc(rs))
+							Routes.CreateMapping(rs.externalContainerName, rs.containerEndpoint, w.makeWakerFunc(rs), w.makeSleeperFunc(rs))
 						} else {
 							Routes.SetDefaultRoute(rs.containerEndpoint)
 						}
--- a/server/k8s.go
+++ b/server/k8s.go
@@ -156,7 +156,7 @@ func (w *k8sWatcherImpl) handleUpdate(oldObj interface{}, newObj interface{}) {
 			"new": newRoutableService,
 		}).Debug("UPDATE")
 		if newRoutableService.externalServiceName != "" {
-			Routes.CreateMapping(newRoutableService.externalServiceName, newRoutableService.containerEndpoint, newRoutableService.autoScaleUp)
+			Routes.CreateMapping(newRoutableService.externalServiceName, newRoutableService.containerEndpoint, newRoutableService.autoScaleUp, newRoutableService.autoScaleDown)
 		} else {
 			Routes.SetDefaultRoute(newRoutableService.containerEndpoint)
 		}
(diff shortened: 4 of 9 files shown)


Sammcb:
@itzg This isn't polished up yet, but wanted to get your thoughts on this feature and the current implementation.

Would you prefer to have this functionality under new options like I currently have, or would you want to condense all the `AutoScaleUp` and `AutoScaleDown` options into a more concise config? Do you want to allow users to enable the up/down auto scalers independently?

On a more technical note, the way I am currently figuring out if a scale down routine should be created is by first checking the `Connector.activeConnections` value. This is fine for now as only one instance of `Connector` is ever created. But the `DownScaler` is a singleton (like `Routes`), so this current logic will be an issue if there are ever plans to have multiple `Connector` instances. Would you prefer I add logic to the `DownScaler` to track number of active connections per sever, or just stick with using the `Connector.activeConnections` value for now?

Really appreciate any additional thoughts/feedback!

itzg:
The singleton of DownScaler seems good to me since it tracks Routes, like you said, and connector.

It does seem like it would be better to track and use active connections per backend. I'm surprised I hadn't already been maintaining a metric for that.

Sammcb:
> It does seem like it would be better to track and use active connections per backend. I'm surprised I hadn't already been maintaining a metric for that.

Where would this be best setup? If `Connector` is a singleton too (which some small code changes might better indicate) then I think it would be best to add there, but if not then would you want some new structure to store that data?

cmd/mc-router/main.go
```diff
@@ -124,6 +126,14 @@ func main() {
 
 	metricsBuilder := NewMetricsBuilder(config.MetricsBackend, &config.MetricsBackendConfig)
 
+	downScalerEnabled := config.AutoScaleDown && (config.InKubeCluster || config.KubeConfig != "")
+	downScalerDelay, err := time.ParseDuration(config.AutoScaleDownAfter)
+	if err != nil {
+		logrus.WithError(err).Error("Unable to parse down scale after duration")
```
alexfrs69, line 132:
```suggestion
		logrus.WithError(err).Error("Unable to parse down scale after duration")
		downScalerDelay = 10 * time.Minute
```
Sorry if I’m misunderstanding, but shouldn’t downScalerDelay be set to the default value if the config is in an incorrect format?

Sammcb: Actually, I was planning to make this `Fatal` so the program exists (I want the user to know they set an incorrect value rather than making it more difficult to figure out by only emitting the log).

itzg:
Yeah, I think the counts would be best maintained/associated with the connector object. I could go either way with it as a separate object.

cmd/mc-router/main.go
itzg: Yeah, I think a fatal would be a choice here.

Sammcb:
Sounds good, I'll go ahead and add it to the `Connector` for now!

Sammcb:
@itzg Ok, I added a new struct to track the per-server network connections and a metric that is emitted. Before I make README updates, would you prefer for the current autoscale config options to be condensed into a single struct (like the metrics config), or do you like them separated as they are currently?

itzg:
> Before I make README updates, would you prefer for the current autoscale config options to be condensed into a single struct (like the metrics config), or do you like them separated as they are currently?

Yeah, it does look like it's time to move those into their own struct. The good news is that if you call the field `AutoScale` in the `Config` struct, then the rendered command line args will remain the same if the fields in the new struct are `Down`, `DownAfter`, etc.

server/connector.go
```diff
@@ -66,7 +67,45 @@ func (p *PlayerInfo) String() string {
...
+	if activeConnections, ok := sm.activeConnections[serverAddress]; ok {
+		return activeConnections
+	}
+	return 0
+}
+
 func NewConnector(metrics *ConnectorMetrics, sendProxyProto bool, receiveProxyProto bool, trustedProxyNets []*net.IPNet, recordLogins bool, autoScaleUpAllowDenyConfig *AllowDenyConfig) *Connector {
+	serverMetrics := &ServerMetrics{
+		activeConnections: make(map[string]int),
+	}
```
itzg, line 107: For consistency, introduce a `NewServerMetrics` "constructor" to wrap up the field init.

cmd/mc-router/main.go
```diff
@@ -124,6 +126,15 @@ func main() {
 
 	metricsBuilder := NewMetricsBuilder(config.MetricsBackend, &config.MetricsBackendConfig)
 
+	downScalerEnabled := config.AutoScaleDown && (config.InKubeCluster || config.KubeConfig != "")
+	downScalerDelay, err := time.ParseDuration(config.AutoScaleDownAfter)
+	if err != nil {
+		logrus.WithError(err).Fatal("Unable to parse down scale after duration")
```
itzg, line 132:
```suggestion
		logrus.WithError(err).Fatal("Unable to parse auto scale down after duration")
```

server/connector.go
```diff
@@ -342,7 +383,10 @@ func (c *Connector) readUserInfo(bufferedReader *bufio.Reader, clientAddr net.Ad
 func (c *Connector) findAndConnectBackend(ctx context.Context, frontendConn net.Conn,
 	clientAddr net.Addr, preReadContent io.Reader, serverAddress string, playerInfo *PlayerInfo, nextState mcproto.State) {
 
-	backendHostPort, resolvedHost, waker := Routes.FindBackendForServerAddress(ctx, serverAddress)
+	// Should always be the last defer statement
```
itzg, line 386: Now that there's three defer's in the same function I'm thinking it would be cleaner to introduce a new 'cleanupBackendConnection` type method that becomes the target of a single defer. One of the things refactored into that function's args would be an indication of whether down scaling should begin.

server/docker.go
```diff
@@ -41,6 +41,12 @@ func (w *dockerWatcherImpl) makeWakerFunc(_ *routableContainer) func(ctx context
 	}
 }
 
+func (w *dockerWatcherImpl) makeSleeperFunc(_ *routableContainer) func(ctx context.Context) error {
+	return func(ctx context.Context) error {
+		return nil
```
itzg, line 46: I'm wondering if both makeWakerFunc and this should be logging a fatal that auto up/down is not supported (yet) for Docker.

server/k8s.go
```diff
@@ -281,3 +283,41 @@ func (w *k8sWatcherImpl) buildScaleUpFunction(service *core.Service) func(ctx co
 		return nil
 	}
 }
+
+func (w *k8sWatcherImpl) buildScaleDownFunction(service *core.Service) func(ctx context.Context) error {
```
itzg, line 287: This and `buildScaleUpFunction` can probably be refactored into a combined method I think. I'm thinking they're main difference would be via `from` and `to` arguments to specify which way the scaling is going.

server/routes.go
```diff
@@ -70,7 +70,7 @@ func routesCreateHandler(writer http.ResponseWriter, request *http.Request) {
 		return
 	}
 
-	Routes.CreateMapping(definition.ServerAddress, definition.Backend, func(ctx context.Context) error { return nil })
+	Routes.CreateMapping(definition.ServerAddress, definition.Backend, func(ctx context.Context) error { return nil }, func(ctx context.Context) error { return nil })
```
itzg, line 73: It's probably time to declare a typedef for a `ScalerFunc` that is `func(ctx context.Context) error`. These growing return declarations will likely be easier to interpret.

Sammcb:
@itzg Ok, made updates to the config input and I think I addressed the comments you left but please let me know if there's anything I missed!

Sammcb:
One breaking change to highlight here is that I did rename the `-auto-scale-up-allow-deny` argument to just `-auto-scale-allow-deny` as the list does affect who can cancel an active down scaler. I'm happy to revert this if you prefer, but I do think the new argument name is a better indicator of the functionality it affects.

itzg:
> One breaking change to highlight here is that I did rename the `-auto-scale-up-allow-deny` argument to just `-auto-scale-allow-deny` as the list does affect who can cancel an active down scaler. I'm happy to revert this if you prefer, but I do think the new argument name is a better indicator of the functionality it affects.

Yep, that sounds good. Since you only just added the argument name, now is the time to tidy that up with little to no impact on users.

README.md
```diff
@@ -267,13 +271,13 @@ kubectl apply -f https://raw.githubusercontent.com/itzg/mc-router/master/docs/k8
 * I extended the allowed node port range by adding `--service-node-port-range=25000-32767`
   to `/etc/kubernetes/manifests/kube-apiserver.yaml`
 
-##### Auto Scale Up
+##### Auto Scale Up/Down
 
-The `-auto-scale-up` flag argument makes the router "wake up" any stopped backend servers, by changing `replicas: 0` to `replicas: 1`.
+The `-auto-scale-up` flag argument makes the router "wake up" any stopped backend servers by changing `replicas: 0` to `replicas: 1`. The `-auto-scale-down` flag argument makes the router shut down any running backend servers with no active connections by changing `replicas: 1` to `replicas: 0`. The scale down will occure after a configurable (using the `-auto-scale-down-after` argument) waiting period, such as `10m` (10 minutes), `2h` (2 hours), etc. If any players connect to the server during this period the scale down will be canceled. It is recommended to set this value high enough so a temporary player disconnect will not immediately shut down the server (`1m` or higher).
```
itzg, line 276:
```suggestion
The `-auto-scale-up` flag argument makes the router "wake up" any stopped backend servers by changing `replicas: 0` to `replicas: 1`. The `-auto-scale-down` flag argument makes the router shut down any running backend servers with no active connections by changing `replicas: 1` to `replicas: 0`. The scale down will occur after a configurable (using the `-auto-scale-down-after` argument) waiting period, such as `10m` (10 minutes), `2h` (2 hours), etc. If any players connect to the server during this period the scale down will be canceled. It is recommended to set this value high enough so a temporary player disconnect will not immediately shut down the server (`1m` or higher).
```

README.md
```diff
@@ -267,13 +271,13 @@ kubectl apply -f https://raw.githubusercontent.com/itzg/mc-router/master/docs/k8
...
 
-This requires using `kind: StatefulSet` instead of `kind: Service` for the Minecraft backend servers.
+Both options requires using `kind: StatefulSet` instead of `kind: Service` for the Minecraft backend servers.
```
itzg, line 278:
```suggestion
Both options require using `kind: StatefulSet` instead of `kind: Service` for the Minecraft backend servers.
```

server/k8s.go
```diff
@@ -239,12 +240,13 @@ func (w *k8sWatcherImpl) buildDetails(service *core.Service, externalServiceName
...
 		containerEndpoint:   net.JoinHostPort(clusterIp, port),
-		autoScaleUp:         w.buildScaleUpFunction(service),
+		autoScaleUp:         w.buildScaleFunction(service, 0, 1),
+		autoScaleDown:       w.buildScaleFunction(service, 1, 0),
 	}
 	return rs
 }
 
-func (w *k8sWatcherImpl) buildScaleUpFunction(service *core.Service) func(ctx context.Context) error {
+func (w *k8sWatcherImpl) buildScaleFunction(service *core.Service, from int, to int) ScalerFunc {
```
itzg, line 249: nit: since you're needing to cast them to int32 for kubernetes API, perhaps just declare as int32 in the method args.

server/k8s.go
```diff
@@ -255,23 +257,23 @@ func (w *k8sWatcherImpl) buildScaleUpFunction(service *core.Service) func(ctx co
...
 						},
-						Spec: autoscaling.ScaleSpec{Replicas: 1}}, meta.UpdateOptions{},
+						Spec: autoscaling.ScaleSpec{Replicas: int32(to)}}, meta.UpdateOptions{},
 					); err == nil {
 						logrus.WithFields(logrus.Fields{
 							"service":     serviceName,
 							"statefulSet": statefulSetName,
 							"replicas":    replicas,
-						}).Info("StatefulSet Replicas Autoscaled from 0 to 1 (wake up)")
+						}).Info("StatefulSet Replicas Autoscaled from "+strconv.Itoa(from)+" to "+strconv.Itoa(to))
```
itzg, line 274: logrus provides formatted versions, so could do

```suggestion
						}).Infof("StatefulSet Replicas Autoscaled from %d to %d", from, to)
```

server/k8s.go
```diff
@@ -255,23 +257,23 @@ func (w *k8sWatcherImpl) buildScaleUpFunction(service *core.Service) func(ctx co
...
 					} else {
-						return errors.Wrap(err, "UpdateScale for Replicas=1 failed for StatefulSet: "+statefulSetName)
+						return errors.Wrap(err, "UpdateScale for Replicas="+strconv.Itoa(to)+" failed for StatefulSet: "+statefulSetName)
```
itzg, line 276: and same here

server/routes.go
```diff
@@ -12,6 +12,12 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+type ScalerFunc func(ctx context.Context) error
+
+func EmptyScalerFunc() ScalerFunc {
+	return func(ctx context.Context) error { return nil }
+}
```
itzg, line 19: Could simplify further with

```suggestion
var EmptyScalerFunc = func(ctx context.Context) error {return nil}
```

server/routes.go
```diff
@@ -70,7 +76,7 @@ func routesCreateHandler(writer http.ResponseWriter, request *http.Request) {
 		return
 	}
 
-	Routes.CreateMapping(definition.ServerAddress, definition.Backend, func(ctx context.Context) error { return nil })
+	Routes.CreateMapping(definition.ServerAddress, definition.Backend, EmptyScalerFunc(), EmptyScalerFunc())
```
itzg, line 79: and

```suggestion
	Routes.CreateMapping(definition.ServerAddress, definition.Backend, EmptyScalerFunc, EmptyScalerFunc)
```

itzg (commented):
Awesome. Just a few minor things left.

itzg:
I was going to do a sanity test as soon as I get a chance, but wanted to double check that you've tested the changes with and without the auto scale up/down enabled?

Sammcb:
Yes, I have tested with both options enabled and disabled in a development Kubernetes cluster. Seemed to perform as expected. But actually now that I think about it would the change to add `Fatal` logs to the wake/sleep functions for docker/swarm cause issues? I think those default functions would mean the `waker != nil` check in the connector would never be false and the functions would always be triggered and result in fatal errors right?

Sammcb:
@itzg Just pushed up another commit that should address those issues. Now the Kubernetes, Docker, and Docker Swarm watchers only return scaler functions if the appropriate auto scale option is enabled and otherwise return `nil`. I left the fatal logs in the functions being returned in Docker and Docker Swarm because now those should only be returned if the user enables those options. Additionally, with the old logic the Kubernetes auto scale down would have only worked if auto scale up was also enabled so that has been fixed.

I tested this all in my Kubernetes environment with both up and down false, down true and up false, up true and down false, and up true and down true. Everything behaves as expected.

itzg:
I'll take a look at that change, but from what you're describing that sounds like a definite change for the better.

itzg (approved):
Latest updates are great

itzg:
Now released in https://github.com/itzg/mc-router/releases/tag/1.29.0

Issue #36, opened 2021-12-28
Return "last known" server status (at least for backends that are currently unreachable but which have "wake-up" available)


vorburger:
With the introduction of #29 for #21, a possible future enhancement on that front, for anyone interested in contributing here, is this:

Now that `mc-router` can "wake up" back-ends, the "status" in the Minecraft client, where one can see the Minecraft version and MOTD and Player Count and Server Icon and what not, obviously still just shows _"Can't connect to server"_ for backends that are not yet running, but which one CAN actually connect to (upon which they are woken up by increasing the `replicas` from 0 to 1, as per that feature).

In a perfect ideal world, it would be cool if we would e.g. "remember" (cache) the "last known" status response - and return that instead of the _"Can't connect to server"._

This is likely a bit of (fun, interesting) work, because `mc-router` currently just "pipes through" connections - it doesn't actually know about the Minecraft protocol with status responses and such. (#3 has some WIP work about inspecting frames for commands; this is somewhat related to that.) https://github.com/haveachin/infrared#response-status does this kind of thing, so it seems like it's doable in principle. (But I don't like how there you have to specify that status response in a separate configuration; it seems to me that it would be much nicer if we simply "cached" it from the real back-end.)

I'm unlikely going to work on this myself in the near term, but thought opening an issue about it anyway could perhaps motivate someone else reading along to dig into this.

PR #406, opened 2025-04-29
feat(auto-scale-up): add fake online, server status cache, and seamless connection during scale-up

alexfrs69:
This PR introduces a feature to display a configurable "fake online" MOTD when a backend server is offline and auto-scale-up is enabled. When a status ping is received and the backend is not available, the router responds with a custom MOTD instead of appearing offline.
It also "hang" the user while retrying to connect to the backend until it reach a defined timeout and then close the connection.
Overall improving the user experience during backend startup.

Related: #36 #21

## Key Changes
**For fake online:**
- Adds configuration options for enabling/disabling the fake online MOTD and customizing its text.
- Ensures the custom MOTD is only sent if auto-scale-up is enabled for the backend.

**For status requests:**
- Tries to connect to the backend once with a short timeout
- If the backend is unavailable, returns the fake online MOTD.

**For connect requests:**
- If auto-scaler is enabled, the client connection is kept open ("hangs") while retrying backend connection until a configurable timeout is reached.
- If auto-scaler is disabled, only a single backend connection attempt is made.
- Lays the groundwork for further enhancements, such as caching backend status and dynamic MOTD updates.

### Version Response
Currently, the status response uses:
```
Version: StatusVersion{
    Name:     "1.21.5",
    Protocol: 770,
}
```
Should the version and protocol be configurable, or is it fine to hardcode the latest supported version? What version/protocol should we default to for best compatibility?

## Next Steps
- The ultimate goal is to cache backend server status and possibly enhance the MOTD to indicate when a server is sleeping or waking up.


itzg:
Regarding questions:

** Version Response** it would be ideal to ping the backend once at startup and retain its version/protocol; however, in the spirit of incremental PRs I'm fine with it being hardcoded at first. I'll leave it up to you for this initial PR.

alexfrs69:
Thanks for your input @itzg 

It seems that waker() is non-nil even with auto scale up disabled.
Do you think pass the global autoScaleUp config value as a field in `Connector` struct is our best option?

server/connector.go
itzg, line 1: Looks like an indentation change for the whole method? Any chance you could narrow the diff to just the functional changes? I use IntelliJ (aka GoLand)...is it an editor difference; however, I thought Go standardizes on indentation.

itzg:
> It seems that `waker` is non-nil even with auto scale up disabled.
> Do you think pass the global autoScaleUp config value as a field in `Connector` struct is our best option?
> Like the `fakeOnline` config, but we’re starting to pass a lot of individual configuration variables

Yeah, it now looks like it would be good to bundle up the config parameters to avoid the argument/field growth. Since `waker` is always non-nil, are you pondering shifting the new fields into that? If so, that sounds good to me.

server/connector.go
alexfrs69: Should be fixed, my editor suddenly preferred spaces over tabs

alexfrs69:
I'm pretty new to golang. Sorry if the code is a bit messy.
I think that my `time.Sleep(backendRetryInterval)` isn't something good to do as it block the whole main thread?
There is a lot of repeating code, there's much space for improvement

server/connector.go
itzg: Thanks. I don't think I realized that the official gofmt is using tabs this whole time 😄 

> The official recommendation, enforced by the gofmt tool, is to use tabs for indentation and spaces for alignment. This ensures consistency across all Go codebases.

server/connector.go
```diff
@@ -394,24 +402,110 @@ func (c *Connector) findAndConnectBackend(ctx context.Context, frontendConn net.
...
+			}
+			return
+		}
+	} else if nextState == mcproto.StateLogin {
+		// Connect request
+		if waker != nil {
+			logrus.Debug("Connect: Autoscaler is enabled, waiting for backend to be ready")
+			// Autoscaler enabled: retry until backendTimeout is reached
+			deadline := time.Now().Add(backendTimeout)
+			for {
```
itzg, line 453: In the past I have liked and been happy with https://github.com/sethvargo/go-retry where you can use a max duration like you're implementing https://github.com/sethvargo/go-retry?tab=readme-ov-file#withmaxduration.

I'm wondering if wrapping that retry wider, limiting the retry to none (in the case of non-status), might help to simplify down the new logic here.

alexfrs69: Thanks for the advice, seems much cleaner.
Let me know what you think of the current logic.

itzg:
Sorry for the delay getting back to your PR. I'm focusing on #405 and then after that's merged can shift to yours.

alexfrs69:
> Sorry for the delay getting back to your PR. I'm focusing on https://github.com/itzg/mc-router/pull/405 and then after that's merged can shift to yours.

No problem 😃 
I'll work on the refactor of the config in the meantime

server/cache.go
```diff
@@ -0,0 +1,122 @@
...
+	return status, true
+}
+
+func (sc *StatusCache) Set(serverAddress string, status *CachedStatus) {
+	sc.mu.Lock()
+	defer sc.mu.Unlock()
+	sc.cache[serverAddress] = status
+}
+
+func (sc *StatusCache) Delete(serverAddress string) {
```
alexfrs69, line 53: Do we need to delete status cache of server that are not anymore in mapping?

server/cache.go
```diff
@@ -0,0 +1,122 @@
...
+type StatusCache struct {
+	mu    sync.RWMutex
+	cache map[string]*CachedStatus // key: serverAddress
+	ttl   time.Duration
+}
+
+func NewStatusCache(ttl time.Duration) *StatusCache {
+	return &StatusCache{
+		cache: make(map[string]*CachedStatus),
+		ttl:   ttl,
```
alexfrs69, line 33: Do we really need a TTL?

server/cache.go
```diff
@@ -0,0 +1,122 @@
...
+import (
+	"net"
+	"strconv"
+	"sync"
+	"time"
+
+	"github.com/itzg/mc-router/mcproto"
+	"github.com/sirupsen/logrus"
+
+	mcpinger "github.com/Raqbit/mc-pinger"
```
alexfrs69, line 12: More simple to use a library to fetch server status, saw you used that in mc-monitor

server/cache.go
```diff
@@ -0,0 +1,122 @@
...
+// CachedStatus holds the cached status response for a backend.
+type CachedStatus struct {
+	Version     mcproto.StatusVersion
+	Description mcproto.StatusText
+	Favicon     string
+	Players     mcproto.StatusPlayers
+	LastUpdated time.Time
+}
+
+type StatusCache struct {
```
alexfrs69, line 24: heavily inspired from routes mappings, can maybe simplify a little bit by removing `mu` key?

itzg:
Previous PR is now merged, so now I can focus on yours after you get merge conflicts caught up.

alexfrs69:
Conflicts should be fixed @itzg

itzg:
Thanks @alexfrs69 

@Sammcb since you've been in some this code recently I was wondering if you could help look at this PR also?

itzg:
@alexfrs69 just noticed the PR is still in draft. Want to mark it ready for review?

server/connector.go
```diff
@@ -491,12 +542,56 @@ func (c *Connector) findAndConnectBackend(ctx context.Context, frontendConn net.
...
+
+			var status *mcproto.StatusResponse
+
+			if c.config.CacheStatus {
+				cachedStatus, ok := c.StatusCache.Get(serverAddress)
+				if ok {
+					logrus.WithField("cachedStatus", cachedStatus).Debug("Using cached status")
+					status = &mcproto.StatusResponse{
+						Players:     cachedStatus.Players,
+						Description: cachedStatus.Description,
```
alexfrs69, line 562: Currently `FakeOnlineMOTD` is displayed only if we don't have a cached status for the backend.
Would be nice to add an option to join the `Description` with `FakeOnlineMOTD` ?

cmd/mc-router/main.go
```diff
@@ -184,6 +203,15 @@ func main() {
 			server.NewWebhookNotifier(config.Webhook.Url, config.Webhook.RequireUser))
 	}
 
+	var cacheInterval time.Duration
+	if config.CacheStatus {
+		cacheInterval, err = time.ParseDuration(config.CacheStatusInterval)
+		if err != nil {
+			logrus.WithError(err).Fatal("Unable to parse cache status interval")
+		}
+		logrus.WithField("interval", config.CacheStatusInterval).Info("Using cache status interval")
```
Sammcb, line 212:
```suggestion
		logrus.WithField("interval", config.CacheStatusInterval).Debug("Using cache status interval")
```

server/connector.go
```diff
@@ -479,10 +493,54 @@ func (c *Connector) findAndConnectBackend(ctx context.Context, frontendConn net.
 		WithField("player", playerInfo).
 		Info("Connecting to backend")
 
-	backendConn, err := net.Dial("tcp", backendHostPort)
-	if err != nil {
+	// We want to try to connect to the backend every backendRetryInterval
+	var backendTry retry.Backoff
```
Sammcb, line 497: `findAndConnectBackend` is quite a large function now. I'm wondering if it would make sense to move all this new logic into a function like `retryBackendConnection`

server/connector.go
```diff
@@ -491,12 +549,65 @@ func (c *Connector) findAndConnectBackend(ctx context.Context, frontendConn net.
...
 
 		if c.connectionNotifier != nil {
-			notifyErr := c.connectionNotifier.NotifyFailedBackendConnection(ctx, clientAddr, serverAddress, playerInfo, backendHostPort, err)
+			notifyErr := c.connectionNotifier.NotifyFailedBackendConnection(ctx, clientAddr, serverAddress, playerInfo, backendHostPort, retryErr)
 			if notifyErr != nil {
 				logrus.WithError(notifyErr).Warn("failed to notify failed backend connection")
 			}
 		}
 
+		if nextState == mcproto.StateStatus && c.config.FakeOnline && c.config.AutoScaleUp {
```
Sammcb, line 558: Similar to comment above, I'm wondering if it would help the readability to move this into its own function

server/connector.go
```diff
@@ -491,12 +549,65 @@ func (c *Connector) findAndConnectBackend(ctx context.Context, frontendConn net.
...
+			if status == nil {
+				status = &mcproto.StatusResponse{
+					Version:     mcproto.StatusVersion{Name: "1.20.2", Protocol: 770},
+					Players:     mcproto.StatusPlayers{Max: 0, Online: 0},
+					Description: mcproto.StatusText{Text: c.config.FakeOnlineMOTD},
+					Favicon:     "",
+				}
+			}
+
+			writeStatusErr := mcproto.WriteStatusResponse(
```
Sammcb, line 595: Just curious, if this occurs does [this logic](https://github.com/itzg/mc-router/pull/406/files#diff-ec4e0cb1bce7492d87db3f0dd57082fa4ab26229e47baab050850d7e435c866cR693) need to be changed (`amount, err := io.Copy(backendConn, preReadContent)`)? I'm wondering how this would work if the router has already returned as response to the client

server/connector.go
```diff
@@ -491,12 +549,65 @@ func (c *Connector) findAndConnectBackend(ctx context.Context, frontendConn net.
...
+						Favicon:     cachedStatus.Favicon,
+					}
+				} else {
+					logrus.WithField("serverAddress", serverAddress).Debug("No cached status found")
+				}
+			}
+
+			if status == nil {
+				status = &mcproto.StatusResponse{
+					Version:     mcproto.StatusVersion{Name: "1.20.2", Protocol: 770},
```
Sammcb, line 588: I know you discussed this here https://github.com/itzg/mc-router/pull/406#issuecomment-2840459420, but I'm wondering if hard-coding the version to a valid version would be confusing. Does it make sense to return something like `Unknown` instead?

Sammcb: And just curious, would it make more sense to simply ping the server in this instance and add the cache entry here? That way the cache wouldn't need to constantly ping backend servers, it would just happen when a client tries to ping.

Additionally, (apologies if this is a stupid question) but what benefits does caching provide? I'm wondering if it would just be simpler to try and get the status of a server and then modify the returned `Description`. Especially if this is only done when a client tries to get the status, then I don't think this would add any more overhead to the backend servers.

@alexfrs69 @itzg Not sure if this was already discussed, but I'm interested to hear your thoughts!

cmd/mc-router/main.go
```diff
@@ -71,6 +71,12 @@ type Config struct {
 
 	SimplifySRV bool `default:"false" usage:"Simplify fully qualified SRV records for mapping"`
 
+	FakeOnline     bool   `default:"false" usage:"Enable fake online MOTD when backend is offline and auto-scale-up is enabled"`
+	FakeOnlineMOTD string `default:"Server is sleeping\nJoin to wake it up" usage:"Custom MOTD to show when backend is offline and auto-scale-up is enabled"`
```
Sammcb, line 75: Should these be moved under the new `AutoScale` struct since they are only functional if it is enabled? https://github.com/itzg/mc-router/blob/8e34da22351fbe035535b22c1677f977879487ab/cmd/mc-router/main.go#L36

server/connector.go
```diff
@@ -434,8 +448,8 @@ func (c *Connector) findAndConnectBackend(ctx context.Context, frontendConn net.
 		c.cleanupBackendConnection(ctx, clientAddr, serverAddress, playerInfo, backendHostPort, cleanupMetrics, cleanupCheckScaleDown)
 	}()
 
-	if waker != nil && nextState > mcproto.StateStatus {
-		serverAllowsPlayer := c.autoScaleUpAllowDenyConfig.ServerAllowsPlayer(serverAddress, playerInfo)
+	if c.config.AutoScaleUp && waker != nil && nextState > mcproto.StateStatus {
```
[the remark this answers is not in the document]
itzg: @Sammcb when you say "including this", I assume you mean the new `c.config.AutoScaleUp` condition?

Agreed though on your follow up thought -- let's leave it to make it clear the intent.

server/connector.go
itzg: Yeah, it does seem to be time for refactoring; however, I would be ok with saving that for a separate PR to minimize the delta between each maneuver.

server/connector.go
itzg: Now that mc-pinger is in use and given this observation, I'm leaning towards there being at least a one time ping at startup to gather the version and protocol of each backend server....however, without having reviewed all the latest changes that include caching, I'm now confused by that cached info isn't used here?

cmd/mc-router/main.go
itzg: I would prefer that too.

server/connector.go
itzg: ...and I'm already changing my mind from 3 minutes ago. Refactoring this now would be great since I think it'll improve the readability of this logic. At it is, I'm having to twist my brain around too much to think through the state diagram of "if down, but waking and it's server list ping, then retry..."

alexfrs69: I agree, i'll work on refactoring that

server/connector.go
alexfrs69: > I'm wondering if hard-coding the version to a valid version would be confusing. Does it make sense to return something like Unknown instead?

I'll make some test to see what minecraft client display when we return non standard protocol status response

> And just curious, would it make more sense to simply ping the server in this instance and add the cache entry here? That way the cache wouldn't need to constantly ping backend servers, it would just happen when a client tries to ping.

So we loose the feature of displaying cached status when a backend is stopped. The only goal of the cache currently is to return a status that a stopped backend had previously responded

> Additionally, (apologies if this is a stupid question) but what benefits does caching provide? I'm wondering if it would just be simpler to try and get the status of a server and then modify the returned Description. Especially if this is only done when a client tries to get the status, then I don't think this would add any more overhead to the backend servers.

Caching is only used in the context of auto scale up, when a backend is stopped, we can't have the version/protocol backend was using. Caching previous status fix that. I agree that the logic behind caching is a bit dumb and add some overhead to the minecraft servers, we could cache status at the startup and when a client requests it.
(There is no stupid questions, just stupid answer 😄 )

> I'm leaning towards there being at least a one time ping at startup to gather the version and protocol of each backend server

At the startup we request and cache every backend status

> I'm now confused by that cached info isn't used here?

We use it, i'm checking if we have it in cache, if we don't have it, we return a default one

Sammcb: @alexfrs69 Ah, I understand now. Thanks! I know you were asking about needing a TTL and I think for this use case a TTL doesn't make much sense as you would want to persist a cached state for a existent, yet shutdown backend until it is up again.

I do see how the elements of a cached response would be nice for making it look like the server was online. I would override the cached players online though, as that should be 0 when the backend is offline.

alexfrs69: > I would override the cached players online though, as that should be 0 when the backend is offline.

Already done, we return only the maxPlayer

Sammcb: Ah never mind I had missed that, thanks!

server/connector.go
alexfrs69: If we write a status to the client we return just after. I don't think this will be reached

alexfrs69:
Thanks for the reviews @itzg @Sammcb!

I've addressed your requested changes, please take another look 🙂
I haven't tested thoroughly yet, but I noticed an issue:

If a client repeatedly refreshes the server status page, each status request is handled sequentially, causing a backlog and significant response delays. This could potentially lead to a DoS situation. How do you suggest we address this?

Sammcb:
@alexfrs69 So to understand the high level goals. The fake online (and associated caching) and fake online MOTD really only comes into play in the case where a mapping for a backend exists, but that backend is not reachable and the user has auto scale up enabled?

If that is the case, my understanding is the final goal is to capture information about a backend server's state to display in status requests to the player when that server is down. One case I don't think is solved by this is if all backend servers are scaled down when `mc-router` starts up. Unless it wakes up all backend servers, it would not be able to get info to show to players in fake statuses. Additionally, there would be similar cases where a user might take down a backend server to make updates (version upgrades, config changes, etc.)  and then leave it down until a player tries to connect. In this case, these updates might not match the status shown to players.

I'm wondering what your thoughts would be on instead of trying to get all this info from the server, simply adding a config file the user could supply with fallback status info? Obviously, this could drift from the running server status too and would be up to the user running the servers/router to keep things synced.

Thank you for doing so much extensive testing and research on these changes so far! Just wanted to suggest the config option route as I recently ran into this issue when wanting to setup allow/deny lists for server autoscalers. I realized it would be very difficult to get the information from the servers directly so opted to simply allow users to enter it as a config for the `mc-router`. I don't mean to intrude on your use case though, just something I thought of as another approach.

alexfrs69:
> So to understand the high level goals. The fake online (and associated caching) and fake online MOTD really only comes into play in the case where a mapping for a backend exists, but that backend is not reachable and the user has auto scale up enabled?

Exactly

I like the idea! I definitely considered the limitation that mc-router can't get the status if the backend is stopped. Relying only on configuration for status at scale could be a bit heavy to manage, though.

Maybe we could accept this limitation for now, or as you suggested, allow users to provide config files with fallback status info for backends that aren't in the cache. That way, we can return the configured status if we don't have a cached one. This would give flexibility without making config management too laborious for every backend.

Sammcb:
Yeah, I think either accepting that as a limitation or adding the config file input could work!

In terms of getting the data dynamically, I really like this approach (made some slight modifications to your initial suggestion just to clarify):

- Cache all _(available)_ backends at startup
- On each player status request, refresh the cache only if the TTL has expired _or an entry in the cache does not exist_
- When a downscale is triggered, cache the status before setting replicas to 0

alexfrs69:
I can work on implementing the config files tomorrow 🙂

It would also be great to automatically detect any new backends added via k8s labels and cache their status right away.

Sammcb:
If you hook that into the router CreateMapping, etc. then that would be totally doable! (I did something very similar for the down scaler in #405)

cmd/mc-router/main.go
```diff
@@ -184,6 +201,15 @@ func main() {
 			server.NewWebhookNotifier(config.Webhook.Url, config.Webhook.RequireUser))
 	}
 
+	var cacheInterval time.Duration
+	if config.AutoScale.CacheStatus {
+		cacheInterval, err = time.ParseDuration(config.AutoScale.CacheStatusInterval)
```
itzg, line 206: It seems like this config processing and then ultimately `cacheInterval` should be fed into server.ConnectorConfig

cmd/mc-router/main.go
```diff
@@ -240,6 +266,15 @@ func main() {
 		logrus.WithError(err).Fatal("Unable to start metrics reporter")
 	}
 
+	if config.AutoScale.CacheStatus {
+		logrus.Info("Starting status cache updater")
+		connector.StatusCache.StartUpdater(connector, cacheInterval, func() map[string]string {
+			mappings := server.Routes.GetMappings()
+			logrus.WithField("mappings", mappings).Debug("Updating status cache with mappings")
+			return mappings
+		})
```
itzg, line 275: ...and that starting of the cache updater should be encapsulated inside of the `connector.WaitForConnections` method.

As it is, it's leaking a bit too much of runtime behavior into the main method.

server/cache.go
itzg: Ideally, yes. Otherwise it would be a (very, very small) memory leak, right?

server/connector.go
```diff
@@ -625,6 +647,107 @@ func (c *Connector) pumpConnections(ctx context.Context, frontendConn, backendCo
...
+		// If we can't get the status from the cache, send a default status
+		return &mcproto.StatusResponse{
+			Version: mcproto.StatusVersion{
+				Name:     "UNKNOWN",
+				Protocol: 0,
+			},
+			Players: mcproto.StatusPlayers{
+				Max:    0,
+				Online: 0,
+				Sample: []mcproto.PlayerEntry{},
```
itzg, line 716: Seems to be a compile error here. Probably meant StatusPlayerEntry?

server/connector.go
```diff
@@ -6,25 +6,30 @@ import (
...
 	"github.com/juju/ratelimit"
 	"github.com/pires/go-proxyproto"
+	"github.com/sethvargo/go-retry"
 	"github.com/sirupsen/logrus"
 )
 
 const (
-	handshakeTimeout = 5 * time.Second
+	handshakeTimeout     = 5 * time.Second
+	backendTimeout       = 30 * time.Second
```
itzg, line 30: Can  you double check this is less than the timeout that the Minecraft client seems to run when attempting to connect?

server/connector.go
```diff
@@ -625,6 +647,107 @@ func (c *Connector) pumpConnections(ctx context.Context, frontendConn, backendCo
...
+	if !hit {
+		logrus.
+			WithField("serverAddress", serverAddress).
+			Debug("Failed to get status from cache, sending default status")
+
+		// If we can't get the status from the cache, send a default status
+		return &mcproto.StatusResponse{
+			Version: mcproto.StatusVersion{
+				Name:     "UNKNOWN",
+				Protocol: 0,
```
itzg, line 711: Have you been able to see what the Minecraft client thinks of these version and protocol values?

server/cache.go
itzg: Yeah, that would make it consistent. I go back and forth about the unnamed/inheritance style like I used in `routesImpl` 😄

server/cache.go
```diff
@@ -0,0 +1,115 @@
...
+	sc.cache[serverAddress] = status
+}
+
+func (sc *StatusCache) Delete(serverAddress string) {
+	sc.mu.Lock()
+	defer sc.mu.Unlock()
+	delete(sc.cache, serverAddress)
+}
+
+func (sc *StatusCache) updateAll(getBackends func() map[string]string) {
```
itzg, line 48: Nice encapsulation of passing in the getter function! Since it gets passed in a couple of places, I wonder if it could be provided once to `NewStatusCache`?

itzg:
> If a client repeatedly refreshes the server status page, each status request is handled sequentially, causing a backlog and significant response delays. This could potentially lead to a DoS situation. How do you suggest we address this?

That's somewhat by design since `Connector.acceptConnections` runs its accept-loop in a single go routine. I couldn't see in the Go stdlib docs, but I was trying to find where the accept queue length could be specified. In any case, the OS layer I believe should constrain the pending connections to accept, so I think it's safe enough to not explicitly handle.

alexfrs69:
Hey guys, just letting you know that i dont have time to invest in this PR this week.
I'll work on it surely next week 🙂

PR #412, opened 2025-05-11
Feature sighup reload routes

lobo235:
I took a stab at adding SIGHUP handling which will cause routes to be reloaded from disk if a routes config file is being used. I made a change to routes_config.go to allow ReloadRoutesConfig() to be called. This works well to also remove old routes given that it calls the Routes.Reset() function in additional to reloading routes from disk.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,8 @@ The following shows a JSON file for routes config, where `default-server` can al
 }
 ```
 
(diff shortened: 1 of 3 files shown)


itzg (approved):
Thanks! Great job evolving the precious enhancement.

itzg:
This is now released in https://github.com/itzg/mc-router/releases/tag/1.31.0

PR #419, opened 2025-06-20
feat(k8s): ExternalName service support

FedotCompot:
Enables using k8s service discovery, and as consequence allows config live reload, for minecraft servers that are external to the k8s cluster.

Example usecases:
1. there is a separate machine that runs minecraft servers that is in same LAN as k8s cluster
2, the IP of the minecraft server is blocked in the country of some players, use mc-router as jumper point hosted on an allowed IP to bypass the block

### DIFF
--- a/server/k8s.go
+++ b/server/k8s.go
@@ -235,6 +235,9 @@ func (w *k8sWatcherImpl) extractRoutableServices(obj interface{}) []*routableSer
 
 func (w *k8sWatcherImpl) buildDetails(service *core.Service, externalServiceName string) *routableService {
 	clusterIp := service.Spec.ClusterIP
+	if service.Spec.Type == core.ServiceTypeExternalName {
+		clusterIp = service.Spec.ExternalName
+	}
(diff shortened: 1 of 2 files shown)


itzg (commented):
Very cool. Can you update the README also?

README.md
```diff
@@ -253,7 +253,11 @@ metadata:
     "mc-router.itzg.me/externalServerName": "external.host.name,other.host.name"
 ```
 
-mc-router will pick the service port named either `minecraft` or `mc-router`. If neither port names exist, it will use port value 25565.
+### Sevice parsing
```
itzg, line 256:
```suggestion
### Service parsing
```

itzg (commented):
Great update. Just a typo to fix and it's good to merge. 

Thanks!

FedotCompot:
Accepted the change!

itzg:
Released in https://github.com/itzg/mc-router/releases/tag/1.32.0

PR #422, opened 2025-06-29
Fix "\x00FML3\x00" in server address

solidDoWant:
Clients utilizing Forge Mod Loader protocol v3 include `\x00FML3\x00` at the end of the server address in the handshake packet. This shows up in logs, metrics, and webhooks. For example:

```
mc_router_server_active_player{player_name="playerName",player_uuid="00000000-0000-0000-0000-000000000000",server_address="my.server.address�FML3�"} 0
```

This PR truncates everything after the first `\0`.

### DIFF
--- a/mcproto/decode.go
+++ b/mcproto/decode.go
@@ -2,6 +2,8 @@ package mcproto
 
 import (
 	"bytes"
+	"strings"
+
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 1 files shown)


itzg (approved):
Thank! I thought I had fixed that, but must be thinking of something similar.

PR #442, opened 2025-08-13
Ensure CA certs are available for webhooks usage

itzg:
[Discussed in Discord](https://discord.com/channels/660567679458869252/660569641550217327/1405198428211904635)

PR #443, opened 2025-08-13
Use scratch base image but certs from alpine

itzg:
Follow up to #442


itzg:
Need to proceed with merge since there's not yet a fix for https://pkg.go.dev/vuln/GO-2025-3829

Issue #409, opened 2025-05-04
Selectively enable/disable auto-scaling for specific StatefulSets


moritzrinow:
**TLDR**:
Possibility to enable/disable auto-scaling of specific StatefulSets via annotations/labels on the StatefulSet themselves, instead of only globally per mc-router.

**Problem**:
Hi, the StatefulSet auto-scaling feature of mc-router seems really powerful. However if I understood it correctly, it can only be enabled/disabled globally for all server backends.
My use-case involves having some servers automatically scaled down and some not.

**Proposal**:
The first idea that came to my mind was using annotations/labels on the target StatefulSets. It seems simpler and easier to manage than an explicit list configured on mc-router.

itzg:
Thanks for submitting that idea. I agree that an annotation alongside the others makes sense to indicate which ones should or not be considered for auto scaling.

moritzrinow:
@itzg I think this can be closed now.

PR #454, opened 2025-09-08
feat(409): opt-out of auto-scaling per server (k8s)

moritzrinow:
This PR addresses #409 and adds the possibility to selectively disable auto-scaling via annotations on Kubernetes.

Annotations:
- `mc-router.itzg.me/autoScaleUp` for controlling scale-up
- `mc-router.itzg.me/autoScaleDown` for controller scale-down

As mentioned, it is currently only possible to opt-out of auto-scaling, as the configuration flags `--auto-scale-up` and `--auto-scale-down` decide whether or not the actual watcher service is being instantiated or not. However, the piece of logic I added should also work for opt-in if it is being called.

I chose the annotations to work on `Service` rather than `StatefulSet` because the service seems like the main abstraction for the server configuration and the stateful-set is only the backend.

@itzg unfortunately I had to migrate the version of package `golang.ngrok.com/ngrok` down to `v1.12.1` because I was not able to build for `v1.13.0`. It was complaining that this rev does not exist anymore. I checked the Github repo and it does indeed not exist.


moritzrinow:
Technically with this, you can somewhat opt-in to auto-scaling if one of the two (scale-up or scale-down) is enabled via flag.
Example:
One can use the annotation `mc-router.itzg.me/autoScaleUp=true` to opt-in to auto scale-up if the flag `--auto-scale-down` was provided, since the watcher is active then.

itzg (approved):

PR #478, opened 2025-11-19
Revert "Add MC-Router-GUI to Communtiy Solutions"

itzg:
Reverts itzg/mc-router#477

I didn’t realize the other project was a complete fork instead of adding an adjacent web UI service.

### DIFF
--- a/README.md
+++ b/README.md
@@ -597,7 +597,6 @@ In this case the `status` is `"failed-backend-connection"` indicating that a bac
 ## Community Solutions
 
 - **[MC Router Discovery](https://github.com/Seedloaf/mc-router-discovery):** A lightweight sidecar to ensure mc-router is in sync with your list of running servers.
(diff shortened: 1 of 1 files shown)


itzg:
@PhoenixSheppy sorry for discouraging you. It seems this was just a misunderstanding since you said

> was more of a convenience thing while developing, left it in the source though after forgetting to remove it.

If this is only a reaction to my reverting of the link, then I'd rather include the link than know my work is getting spun off into a private project.

itzg:
(I modified my initial comment since I rushed too quickly with my concern of a closed source project)

itzg:
@PhoenixSheppy I still feel really bad for reverting this change. I know this is no excuse and you don't need to reply, but I just want to share that I wasn't in the best mental or physical space last week having to deal with work offsite meetings. I should have waited to review and been more constructive and patient with my concerns.

PR #488, opened 2025-12-09
Docker auto-scale and asleep motd status

Lenart12:


Lenart12:
This is ready for review. I didn't notice #406 is trying to merge about the same functionality.

mcproto/types.go
```diff
@@ -79,6 +79,8 @@ const (
 	PacketIdHandshake            = 0x00
 	PacketIdLogin                = 0x00 // during StateLogin
 	PacketIdLegacyServerListPing = 0xFE
+	PacketIdStatusRequest        = 0x00
+	PacketIdStatusPing           = 0x01
```
itzg: Please call this `PacketIdPingRequest` to match the docs naming convention:

https://minecraft.wiki/w/Java_Edition_protocol/Packets#Ping_Request_(status)

I wasn't able to find a "status ping" when searching the page as is.

mcproto/types.go
```diff
@@ -106,6 +108,11 @@ type LegacyServerListPing struct {
 	ServerPort      uint16
 }
 
+// PingPayload represents the status ping payload (packet 0x01)
+type PingPayload struct {
+	Value int64
```
itzg: This field name should be `Timestamp`, right? According to

https://minecraft.wiki/w/Java_Edition_protocol/Packets#Ping_Request_(status)

mcproto/read.go
```diff
@@ -58,7 +58,30 @@ func ReadPacket(reader *bufio.Reader, addr net.Addr, state State) (*Packet, erro
...
+	// - 0x01 Ping: 8-byte long payload
+	if state == StateStatus {
+		switch packet.PacketID {
+		case PacketIdStatusRequest:
+			// no payload
+			packet.Data = nil
+		case PacketIdStatusPing:
+			// read 8-byte long from remainder
+			if remainder.Len() >= 8 {
+				val := int64(binary.BigEndian.Uint64(remainder.Next(8)))
```
itzg: Use the existing `ReadLong` function in this package

mcproto/write.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+// WriteStatusJSONPacket writes a Status Response (packet 0x00) with the provided JSON string
+func WriteStatusJSONPacket(w io.Writer, jsonString string) error {
+	// payload is the JSON as a Minecraft string
+	var payload bytes.Buffer
+	if err := WriteString(&payload, jsonString); err != nil {
+		return err
+	}
+	pkt := buildPacket(0x00, payload.Bytes())
```
itzg: Declare const for the packet ID `0x00` here

server/docker.go
```diff
@@ -255,13 +369,24 @@ func (w *dockerWatcherImpl) parseContainerData(container *dockertypes.Container)
...
+				logrus.WithFields(logrus.Fields{"containerId": container.ID, "containerNames": container.Name}).
 					Warnf("ignoring container with duplicate %s label", DockerRouterLabelNetwork)
 				return
 			}
 			data.network = new(string)
 			*data.network = value
 		}
+		if key == DockerRouterLabelAutoScaleUp {
+			lowerValue := strings.TrimSpace(strings.ToLower(value))
+			data.autoScaleUp = lowerValue != "" && lowerValue != "0" && lowerValue != "false" && lowerValue != "no"
```
itzg: It would be great to create a helper function for this and then use the same at

https://github.com/itzg/mc-router/blob/22ec39b805bf32e473e21de70fb3b629ae5840dc/server/k8s.go#L287-L291

and below at L385

server/docker.go
```diff
@@ -209,18 +317,24 @@ func (w *dockerWatcherImpl) listContainers(ctx context.Context) ([]*routableCont
...
-	network *string
-	ip      string
+	hosts               []string
+	port                uint64
+	def                 *bool
+	network             *string
+	ip                  string
+	autoScaleDown       bool
+	autoScaleUp         bool
+	autoScaleAsleepMOTD string
```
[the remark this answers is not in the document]
Lenart12: They are parsed because they can be overridden with docker labels. Values from w.config are used as defaults if a value by label is not specified.

server/docker.go
```diff
@@ -59,26 +62,99 @@ func NewDockerWatcher(socket string, timeoutSeconds int, refreshIntervalSeconds
 
 type dockerWatcherImpl struct {
 	sync.RWMutex
-	config dockerWatcherConfig
-	client *client.Client
+	config           dockerWatcherConfig
+	client           *client.Client
+	dockerRoutesLock sync.RWMutex
```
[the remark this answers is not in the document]
Lenart12: Leftover from initial development... whoops

PR #500, opened 2025-12-23
dev builds: Modify Dockerfile to add cache volumes

JLSchuler99:
Adds `--mount=type=cache` for the `go mod download` and `go build` steps.

Adds at least a 10x speedup compared to building from a fresh layer every time.


itzg (approved):
Thanks!

PR #548, opened 2026-05-04
Convert docker polling to event listening

Caedis:
Deprecated the `--docker-refresh-interval`/`DOCKER_REFRESH_INTERVAL`. Value is ignored and a warning is printed for the user.

Any stopped containers that have a `mc-router.host` label will be detected on startup

All docker changes are now instant

### DIFF
--- a/server/docker.go
+++ b/server/docker.go
@@ -9,7 +9,10 @@ import (
 	"sync"
 	"time"
 
+	cerrdefs "github.com/containerd/errdefs"
 	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/api/types/events"
+	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/client"
 	"github.com/sirupsen/logrus"
 )
@@ -30,12 +33,11 @@ const (
 )
 
 type dockerWatcherConfig struct {
-	autoScaleUp     bool
-	autoScaleDown   bool
-	socket          string
-	timeout         time.Duration
-	refreshInterval time.Duration
-	apiVersion      string
+	autoScaleUp   bool
+	autoScaleDown bool
+	socket        string
+	timeout       time.Duration
+	apiVersion    string
 }
 
 func (c *dockerWatcherConfig) apiVersionOpt() client.Opt {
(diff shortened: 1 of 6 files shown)


Caedis:
Could use some testing, i'll swap it in next time my servers are idle

itzg:
FYI, I haven’t forgotten about this.  Just busy around the house this week and want to look over this one carefully.

Caedis:
All good, I have been running it in my server

server/docker.go
```diff
@@ -184,49 +180,175 @@ func (w *dockerWatcherImpl) monitorContainers(ctx context.Context) error {
...
+
+// applyEvent reacts to a single Docker event by reconciling only the routes
+// belonging to the affected container — no full re-list.
+func (w *dockerWatcherImpl) applyEvent(ctx context.Context, ev events.Message) error {
+	containerID := ev.Actor.ID
+	if ev.Type == events.NetworkEventType {
+		containerID = ev.Actor.Attributes["container"]
+	}
+	if containerID == "" {
+		return nil
```
itzg, line 217: Maybe a warning log? Or is this fairly normal?

server/docker.go
```diff
@@ -184,49 +180,175 @@ func (w *dockerWatcherImpl) monitorContainers(ctx context.Context) error {
...
-			} else {
-				Routes.SetDefaultRoute("", "", nil, nil, "", "")
-			}
-			logrus.WithField("routableContainer", rs).Debug("DELETE")
+		if oldRs.containerEndpoint == rs.containerEndpoint &&
+			oldRs.containerID == rs.containerID &&
+			oldRs.autoScaleUp == rs.autoScaleUp &&
+			oldRs.autoScaleDown == rs.autoScaleDown &&
+			oldRs.autoScaleAsleepMOTD == rs.autoScaleAsleepMOTD &&
+			oldRs.autoScaleLoadingMOTD == rs.autoScaleLoadingMOTD {
```
itzg, line 339: An Equals method on routableContainer is tempting, but it would have to exclude externalContainerName

itzg (commented):
Looks great. Just the one question about about a warning log.

server/docker.go
Caedis: That can only really happen if its a daemon-level network event which does not have a related container. Added warning

server/docker.go
Caedis: The container name is the key used for lookup, not a data field. Could give false negatives when a name changes

Caedis: oh, that is what you meant. If this same equal check was used in more than 1 place, i would agree, but its only used here

itzg: Yeah, since it's only once it was just a curiosity 😄

itzg:
Released in https://github.com/itzg/mc-router/releases/tag/v1.42.1

PR #550, opened 2026-05-05
Add timezone support

Caedis:
Supports TZ env or just bind mounting `/etc/localtime`

image


Dockerfile
itzg, line 1: Can you do the same for `Dockerfile.release` since that's what is used for the published image?

itzg (changes requested):
Thanks. Having started out as an exe only image, I had taken those things for granted.

itzg (approved):

itzg:
Build failure is known vuln that doesn't have a published fix yet

PR #553, opened 2026-06-03
Add webhook-based auto-scaling for static routes

lucsoft:
**Why / my requirements:**
1. Scale backends from a separate control process (its own API), not Docker/k8s discovery.
2. Keep mc-router itself minimal & least-privilege — no Docker socket, no kube creds, no shell (distroless-friendly).
3. Need scale-to-zero for static routes, with the scaling decision delegated out of the router.

**What**: First connection → POST scale-up URL; after idle → scale-down URL. An external receiver does the start/stop.

**Design**:  static routes get the global webhook scaler attached at creation — across --mapping, --default, config file, and the REST API, so every static route autoscales the same way.

**My Choices/insights:**

- Webhook, not exec — keeps the scaling privilege (and the shell) out of a network-facing proxy.
- Client can't shape the call: serverAddress/backend come only from config.
- Dynamic backends: scale-up response may return {"backend":"host:port"} to override the address for that wake (e.g. fresh container IP); empty body keeps the configured one. (Kinda required by my example which i put in the docs)
- Global-only by design

**Tests/example**: server/webhook_scaler_test.go, docs/webhook-scaler/ runnable receiver driving itzg/minecraft-server via Apple's container.

image

> Auto-scaling via mc-router. Using Apple's container cli.

### DIFF
--- a/server/configs.go
+++ b/server/configs.go
@@ -22,11 +22,10 @@ type AutoScale struct {
 // POSTing to an external HTTP receiver, which owns the start/stop of the
 // backend. This keeps the scaling privilege out of the router process.
 type AutoScaleWebhookConfig struct {
-	UpUrl       string        `usage:"If set, a POST request is sent to this URL to scale a statically-configured backend up on access"`
-	DownUrl     string        `usage:"If set, a POST request is sent to this URL to scale a statically-configured backend down after idle"`
-	Headers     []string      `usage:"Zero or more 'key=value' headers added to scaler webhook requests, e.g. for authentication tokens"`
-	Timeout     time.Duration `default:"30s" usage:"Timeout for each scaler webhook request"`
-	WakeTimeout time.Duration `default:"60s" usage:"Maximum time to wait for the backend to become reachable after a scale-up webhook"`
+	Url         string            `usage:"If set, statically-configured backends are scaled up on access and down after idle by POSTing to this URL"`
+	Headers     map[string]string `usage:"Zero or more 'key=value' headers added to scaler webhook requests, e.g. for authentication tokens"`
+	Timeout     time.Duration     `default:"30s" usage:"Timeout for each scaler webhook request"`
+	WakeTimeout time.Duration     `default:"60s" usage:"Maximum time to wait for the backend to become reachable after a scale-up webhook"`
 }
 
 type RoutesConfig struct {
--- a/server/configs.go
+++ b/server/configs.go
@@ -3,8 +3,9 @@ package server
 import "time"
 
 type WebhookConfig struct {
-	Url         string `usage:"If set, a POST request that contains connection status notifications will be sent to this HTTP address"`
-	RequireUser bool   `default:"false" usage:"Indicates if the webhook will only be called if a user is connecting rather than just server list/ping"`
+	Url         string        `usage:"If set, a POST request that contains connection status notifications will be sent to this HTTP address"`
+	RequireUser bool          `default:"false" usage:"Indicates if the webhook will only be called if a user is connecting rather than just server list/ping"`
+	Timeout     time.Duration `default:"30s" usage:"Timeout for each connection status notification request"`
 }
 
 type AutoScale struct {
@@ -14,6 +15,18 @@ type AutoScale struct {
 	AllowDeny   string        `usage:"Path to config for server allowlists and denylists. If a global/server entry is specified, only players allowed to connect to the server will be able to trigger a scale up when -auto-scale-up is enabled or cancel active down scalers when -auto-scale-down is enabled"`
 	AsleepMOTD  string        `usage:"MOTD to display when auto-scaled down servers are accessed; if empty, no status will be served"`
 	LoadingMOTD string        `usage:"MOTD to display while auto-scaled Docker servers are waking up; if empty, asleep status will be served"`
+	Webhook     AutoScaleWebhookConfig
+}
+
+// AutoScaleWebhookConfig enables scaling of statically-configured routes by
+// POSTing to an external HTTP receiver, which owns the start/stop of the
+// backend. This keeps the scaling privilege out of the router process.
+type AutoScaleWebhookConfig struct {
+	UpUrl       string        `usage:"If set, a POST request is sent to this URL to scale a statically-configured backend up on access"`
+	DownUrl     string        `usage:"If set, a POST request is sent to this URL to scale a statically-configured backend down after idle"`
+	Headers     []string      `usage:"Zero or more 'key=value' headers added to scaler webhook requests, e.g. for authentication tokens"`
+	Timeout     time.Duration `default:"30s" usage:"Timeout for each scaler webhook request"`
+	WakeTimeout time.Duration `default:"60s" usage:"Maximum time to wait for the backend to become reachable after a scale-up webhook"`
 }
 
 type RoutesConfig struct {
--- a/server/server.go
+++ b/server/server.go
@@ -48,34 +48,30 @@ func NewServer(ctx context.Context, config *Config) (*Server, error) {
 
 	metricsBuilder := NewMetricsBuilder(config.MetricsBackend, &config.MetricsBackendConfig)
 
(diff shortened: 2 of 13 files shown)


itzg:
This sounds interesting, but I'm wondering why not make use of the existing webhooks and add a state change event, if needed? 

Since this design came without prior discussion it will take me a while to review this one.

lucsoft:
@itzg, yeah sorry for not opening an issue. If you have more questions or something we could also move this to a issue if that helps.

I have seen the current webhook_notifier system more like observe-only system as this is more of a hook then a notification. This patch is more of a dynamic way to do scaling, kinda like a hook for the user to do there scaling.

This webhook allows a way to externalize the scaling, yet its still in the same control flow of the process. So if external scaling fails (like 500 error) the process knowns that scaling has failed, so it can react on it.

I assumed if like a webhook fails in webhook_notifier its seemed as fine, its just a notification not a control flow.

itzg:
My hesitation here is around maintaining two parallel webhook systems in the same codebase — duplicated HTTP client setup, retry logic, configuration surface, and documentation.

Before going further I want to dig into the failure handling claim. When the scale-up webhook returns a 500, what does mc-router actually do? As far as I can tell the realistic options are:

- **Refuse the connection** — tell the client the backend is unavailable. This is genuinely useful.
- **Retry with backoff** — but only if that logic is explicitly implemented here.
- **Proceed to the static backend anyway** — which defeats scale-to-zero.
- **Log and give up** — which is functionally fire-and-forget, just synchronous.

If the intent is option 1, that's a meaningful and well-defined behaviour difference from the existing notifier. But there's a deeper race condition regardless: a 200 from the scale-up webhook means "I've been asked to start," not "the backend is ready." So mc-router still needs retry/backoff on the backend TCP connection itself. How is that handled?

On the duplication concern: would it be possible to share the underlying transport infrastructure (HTTP client, timeout config, retry logic) between this and the existing notifier, even if the call semantics stay separate — one async/notify, one synchronous/control-flow with a typed response body? That would address my maintainability concern without collapsing two conceptually different things into one confusing API.

lucsoft:
@itzg 

Thanks for the questions :D 

**Duplication**

- Totally agree, and I'd frame it just like you did: share the transport, keep the semantics separate.
- There's actually no retry logic to merge — neither side retries. It's just the http.Client setup, marshal→POST→close, and headers that overlap.
- Plan: one small shared helper. Notifier keeps its async fire-and-forget wrapper, scaler keeps its sync path with the typed response.
- Some extra: the notifier's hardcoded 30s timeout finally becomes configurable too.

**500 from the scale-up webhook**

- It's option 1 (refuse) — and not option 3. Any >=400 errors out and we close without ever dialing the static backend, so scale-to-zero holds.
- No webhook retry, so not option 2 either.
- Honest caveat: on any wake failure (webhook, Docker, or k8s) we refuse by just dropping the TCP connection rather than sending a real kick packet — so on login the player sees a generic failure instead of "server's starting." That's pre-existing connector behaviour, not new here. Could be addressed in a follow up PR.

**The 200 ≠ ready race**

- Handled ;). the waker doesn't return on the 200. It blocks in waitForBackendReachable, polling the backend every 500ms until it's up or wakeTimeout hits, then refuses if it never comes.
- Two notes: the poll's a fixed 500ms (not exponential), and the connector's final dial is a single attempt leaning on that probe.
- Happy to add backoff if you'd prefer.

If the shared-transport split sounds good, I'll push it as a follow-up commit. Want me to go ahead?

itzg:
Yes, proceed with that plan. I would to see that line count come down.

lucsoft:
@itzg Made these changes now. Had the benefit that webhook_notifier now also uses structured logging as the same code base.

Keep in mind i didn't merge the timeout configurations so notify can have quick timeouts and the scaler webhook can have longer timeouts when the infra takes longer to complete

itzg:
Quick apology, I hadn't looked before (since only on phone so far) and didn't notice the bulk of lines was the docs and sample scaler script.

server/configs.go
```diff
@@ -14,6 +15,18 @@ type AutoScale struct {
...
 	LoadingMOTD string        `usage:"MOTD to display while auto-scaled Docker servers are waking up; if empty, asleep status will be served"`
+	Webhook     AutoScaleWebhookConfig
+}
+
+// AutoScaleWebhookConfig enables scaling of statically-configured routes by
+// POSTing to an external HTTP receiver, which owns the start/stop of the
+// backend. This keeps the scaling privilege out of the router process.
+type AutoScaleWebhookConfig struct {
+	UpUrl       string        `usage:"If set, a POST request is sent to this URL to scale a statically-configured backend up on access"`
+	DownUrl     string        `usage:"If set, a POST request is sent to this URL to scale a statically-configured backend down after idle"`
```
itzg: I'm wondering if these can be combined into a single webhook URL with `action` serving as the discriminator. That would make it consistent with the notifier webhook.

server/server.go
```diff
@@ -48,11 +48,33 @@ func NewServer(ctx context.Context, config *Config) (*Server, error) {
...
+			config.AutoScale.Webhook.UpUrl,
+			config.AutoScale.Webhook.DownUrl,
+			config.AutoScale.Webhook.Headers,
+			config.AutoScale.Webhook.Timeout,
+			config.AutoScale.Webhook.WakeTimeout,
+		)
+		if err != nil {
+			return nil, fmt.Errorf("could not create webhook scaler: %w", err)
+		}
+		WebhookAutoScaler = scaler
```
itzg: I know a lot of the existing code uses a global variable approach, but I'm trying to gradually fix the code to not use that style. Being new code it would be great if you could avoid a global variable and instead pass this instance to those objects that need it.

server/webhook_scaler.go
```diff
@@ -0,0 +1,194 @@
...
+	if err := json.Unmarshal(data, &parsed); err != nil {
+		logrus.WithError(err).Debug("Ignoring unparseable scaler response body")
+		return ""
+	}
+	return strings.TrimSpace(parsed.Backend)
+}
+
+// waitForBackendReachable blocks until a TCP connection to endpoint succeeds,
+// the context is cancelled, or timeout elapses.
+func waitForBackendReachable(ctx context.Context, endpoint string, timeout time.Duration) error {
```
itzg: This needs to be scoped to WebhookScaler. If I were wanting to make use of nested packages then all of these standalone functions could have gone there.

server/webhook_scaler.go
```diff
@@ -0,0 +1,194 @@
...
+	defer resp.Body.Close()
+
+	return parseScaleResponseBackend(resp.Body), nil
+}
+
+// parseScaleResponseBackend extracts an optional backend address from a scaler
+// response body. It is lenient: an empty body, a non-JSON body, or JSON without
+// a "backend" field all yield an empty string, so receivers that return nothing
+// (or a plain "OK") keep working unchanged.
+func parseScaleResponseBackend(body io.Reader) string {
```
itzg: This needs to be scoped to WebhookScaler. If I were wanting to make use of nested packages then all of these standalone functions could have gone there.

server/configs.go
```diff
@@ -14,6 +15,18 @@ type AutoScale struct {
...
+	Headers     []string      `usage:"Zero or more 'key=value' headers added to scaler webhook requests, e.g. for authentication tokens"`
```
itzg: flagsfiller supports map types, so please use that to keep the parsing consistent and also don't need to implement it in parseScalerHeaders

itzg (changes requested):
I still have some more files to review, but wanted to get a first pass back to you sooner.

itzg (commented):
New changes look really good. I still need to review the docs, but I need to do that on a proper screen.

itzg (approved):
Thanks for iterating on the requested changes. Fantastic feature.

itzg:
I'll address the vuln updates in a follow up.

itzg:
Released in https://github.com/itzg/mc-router/releases/tag/v1.43.0

PR #570, opened 2026-07-12
feat: make the backend dial timeout configurable (--backend-dial-timeout)

wachirachoomsiri:
Follow-up to #569, which bounded the backend dial with a fixed 2s timeout so the auto-scale asleep MOTD is served promptly instead of hanging on the OS SYN-retry for a scaled-to-zero backend. This makes that bound configurable via a new `--backend-dial-timeout` flag (env `BACKEND_DIAL_TIMEOUT`), keeping the existing **2s** as the default — so behavior is unchanged unless the flag is set.

### Why

The ideal value depends on how quickly backends are reachable:

- On Kubernetes / in-cluster Services a healthy backend accepts the TCP connection in sub-millisecond time, so operators can lower the timeout (e.g. `100ms`) to serve the asleep MOTD and trigger scale-up almost instantly on a scaled-to-zero Service.
- Deployments with slower or higher-latency backends can keep `2s` (or raise it) to avoid treating a reachable-but-slow backend as asleep.

A single hard-coded constant can't serve both. #569 deliberately chose a conservative 2s; this leaves that as the default while letting operators tune it. (We run mc-router at `100ms` on a Kubernetes Minecraft-hosting deployment where every backend is an on-cluster Service.)

### Changes

- `server/configs.go` — add `BackendDialTimeout time.Duration` (`default:"2s"`) to `Config`.
- `server/connector.go` — replace the `backendDialTimeout` constant with a per-`Connector` field (default `defaultBackendDialTimeout` = 2s) plus a `UseBackendDialTimeout()` setter, matching the existing `UseAsleepMOTD` / `UseClientFilter` style; the dial in `findAndConnectBackend` reads the field.
- `server/server.go` — wire `config.BackendDialTimeout` into the connector.
- `README.md` — document the new flag in the Usage list.

### Testing

- `go vet ./...` and `go test ./server/... ./mcproto/...` pass.
- With no flag set, the dial uses the same 2s bound as before (`defaultBackendDialTimeout`).
- `--help` lists `-backend-dial-timeout duration … (default 2s)`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,8 @@ Some other features included:
     	If set, statically-configured backends are scaled up on access and down after idle by POSTing to this URL (env AUTO_SCALE_WEBHOOK_URL)
   -auto-scale-webhook-wake-timeout duration
     	Maximum time to wait for the backend to become reachable after a scale-up webhook (env AUTO_SCALE_WEBHOOK_WAKE_TIMEOUT) (default 1m0s)
+  -backend-dial-timeout duration
+    	Timeout for establishing the TCP connection to a backend. Bounds how long a dial to a scaled-to-zero server's Service (no endpoints) waits before the auto-scale asleep MOTD / scale-up fallback is served; operators with fast on-cluster backends can lower it so that fallback fires sooner (env BACKEND_DIAL_TIMEOUT) (default 2s)
   -clients-to-allow value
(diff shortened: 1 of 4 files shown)


itzg:
Released in https://github.com/itzg/mc-router/releases/tag/v1.44.1

PR #586, opened 2026-08-02
Various code refactoring and improvements

### LINKED ISSUE #515: Cleanup sleepers/wakers to encapsulate their scaling target
I'm also uncertain about this code in `routesImpl.GetSleepers` where several `SleeperFunc`s are obtained. It seems like that would cause the same scaling operation to be performed multiple times. The outcome is technically correct, but not optimal when a several hostnames map to the same server instance's service.

This was identified as part of #512

itzg:
- correctly handle down scaling when a docker service has multiple external hosts
- introduce ScalingTarget interface with types for docker, swarm, k8s, and webhook discovery/scaling
- clean up code for scale down timer
- avoid race conditions like scale down timer start during wake
- better k8s and docker object logging
- use cgr.dev/chainguard/static as base image
  - **NOTE** the image now runs mc-router as a non-root user by default
- Use avast/retry-go for k8s backend readiness check
- Fix skaffold image reference
- Document manual docker test case (pending a simulated test client)
- Better debug logs in general
- Modern Go cleanup

TODO
- [x] Testing on Swarm
- [x] Testing of webhook scaler

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -47,7 +47,7 @@ dockers_v2:
       - linux/arm/v6
     tags:
       - "{{ .Version }}"
-      - "latest"
+      - "{{ if not .Prerelease }}latest{{ end }}"
     labels:
       org.opencontainers.image.authors: "Geoff Bourne <itzgeoff@gmail.com>"
       org.opencontainers.image.title: "mc-router"
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -18,13 +18,17 @@ Start mc-router directly in the devcontainer.
 
 Verified with skaffold v2.23.0
 
-```
+```shell
 skaffold dev --kube-context=docker-desktop --default-repo=gcr.io/YOURS
 ```
 
 - `YOURS` with your github username or the repo entirely
 
-Also be sure to kubectl apply the minecraft deployment such as `docs/k8s-mc-with-default.yaml`
+Also be sure to kubectl apply the minecraft deployment such as `docs/k8s-mc-deployment.yaml` or `docs/k8s-mc-sts.yaml`, if using autoscaling.
+
+> [!NOTE]
+> The dev image build performed by skaffold uses [ko](https://skaffold.dev/docs/builders/builder-types/ko/), which 
+> does not use the `Dockerfile` in the repo; however, the base image is [the same](https://github.com/chainguard-images/images/tree/main/images/static).
 
 ## Manual test cases
 
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -119,3 +119,36 @@ When applying the `mc-router.host` label to containers to be auto-discovered, it
 
 Run one of the labeled services by clicking the run icon in the gutter.
 
+## Docker Swarm
+
+In order to make development images available to the mc-router service, you need to run an in-cluster retry and push to it.
+
+Start an in-cluster retry service:
+
+```shell
+docker service create \
+  --name registry \
+  --publish published=5000,target=5000 \
+  registry:2
+```
+
+Build the image like normal and tag it as `localhost:5000/itzg/mc-router:r1` using the tag as a distinct version to test.
+
+```shell
+docker tag itzg/mc-router:latest localhost:5000/itzg/mc-router:r1
+```
+
+Finally, push the image to the in-cluster retry service:
+
+```shell
+docker push localhost:5000/itzg/mc-router:r1
+```
+
+Be sure to update the `image` in the stack's compose file, such as
+
+```yaml
+  router:
+     image: 127.0.0.1:5000/itzg/mc-router:r1
+     environment:
+        IN_DOCKER_SWARM: "true"
+```
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -25,3 +25,22 @@ skaffold dev --kube-context=docker-desktop --default-repo=gcr.io/YOURS
 - `YOURS` with your github username or the repo entirely
 
 Also be sure to kubectl apply the minecraft deployment such as `docs/k8s-mc-with-default.yaml`
+
+## Manual test cases
+
+### Docker discovery with auto-scaling
+
+1. Build the image given `Dockerfile` and tag it as `itzg/mc-router`
+2. Start examples/docker-autoscale-multi/compose.yml
+3. Wait for initial scale down of vanilla container
+4. With Minecraft client, do a ping/list and ensure asleep MOTD comes back
+   - Nothing happens with scaling or timer
+5. Connect
+   - Watch for scaling up of vanilla container
+   - Client should fully connect after vanilla container is up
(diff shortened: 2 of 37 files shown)


itzg:
@lucsoft with this PR I am finally doing some code cleanup that I wanted for years (?) now. In the process I updated some of the code paths for webhook based auto-scaling. I was wondering if you could briefly test version `1.46.0-rc1` of the image?

itzg:
@SCBionicle same as above, I was wondering if you could double check Docker Swarm usage with version `1.46.0-rc1` of the image?

> [!IMPORTANT]
> The image runs mc-router as non-root now, so you might need to adjust the user/group as described in https://github.com/itzg/mc-router/tree/feat/scaling-target#user-access-to-docker-socket

lucsoft:
@itzg got a nil pointer exception:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x11e68cb]
goroutine 27 [running]:
sync/atomic.(*Bool).CompareAndSwap(...)
	/opt/hostedtoolcache/go/1.26.5/x64/src/sync/atomic/type.go:29
github.com/itzg/mc-router/server.(*ScalingIndicator).StartScaling(...)
	/home/runner/work/mc-router/mc-router/server/scaling_target.go:26
github.com/itzg/mc-router/server.(*Connector).findAndConnectBackend(0x19a395418000, {0x1663f78, 0x19a3953a00a8}, {0x16513b0, 0x19a3951afce0}, {0x164b640, 0x19a3951afdd0}, {0x19a39522af90, 0xf}, 0x19a395202b60, ...)
	/home/runner/work/mc-router/mc-router/server/connector.go:612 +0x598
github.com/itzg/mc-router/server.(*Connector).HandleConnection(0x19a395418000, {0x1663f78, 0x19a3953a00a8})
	/home/runner/work/mc-router/mc-router/server/connector.go:369 +0x1a68
created by github.com/itzg/mc-router/server.(*Connector).acceptConnections in goroutine 69
	/home/runner/work/mc-router/mc-router/server/connector.go:260 +0x26f

itzg:
Thanks @lucsoft , I'll chase that down. Oh, is there a way I could recreate your full scaling solution  or is it internal code?

lucsoft:
@itzg my cluster IaC is fully open source at https://github.com/lucsoft/cluster 

The important commit: https://github.com/lucsoft/cluster/commit/e287923a24d884a19f7fe561701ca1bfe1d45b4d

SCBionicle:
I have yet to deploy the update, but it is worth mentioning that I have docker community edition installed on my two machines, the Docker Group ID are not the same between nodes on the socket. While that wouldn't affect me as I only have one manager, this could affect the usability on larger swarm clusters to remove root access from the container, if MC-router is allowed to float between nodes.

itzg:
@SCBionicle good point, but to clarify you can always explicitly set the container to run as root. That happened to be the default previously, but I'm now using a more hardened base image. So, it's now user's discretion to raise privileges.

itzg:
> I have docker community edition installed on my two machines, the Docker Group ID are not the same between nodes on the socket

@SCBionicle now I'm wondering why docker would install with a different group ID each time...unless all it does is allocate the next available system user/group IDs. Actually, it turns out you can specify user/group names instead of IDs like `user: "username:groupname"`

Refer to https://docs.docker.com/reference/dockerfile/#user

itzg:
@lucsoft when you get a chance, can you next try `1.46.0-rc2`? Thank you so much. If this one breaks I'll put in the effort to spin up an equivalent test bed 😄 .

itzg:
BTW, @SCBionicle, the webhook autoscaler feature was added by @lucsoft to avoid the elevated user access needed for the docker socket and k8s endpoint 😉

lucsoft:
@itzg can confirm 1.46.0-rc2 works

And yeah i didn't like that mc-router had kube api access. so having that networking boundary to reduce the attack surface is better IMO and well it allows for other api bindings to, like apple container cli like in the example so quite nice haha

itzg:
Thanks @lucsoft !

itzg:
@SCBionicle please refer to this updated example to see a robust and secure way to access the docker socket with swarm mode

https://github.com/itzg/mc-router/blob/208fb59212285e3a906943c07b1f41772b6d76b1/examples/swarm/compose.yml#L2-L19

SCBionicle:
Thank you for the example. That'll make it easier to try out this change. I'll post back here once after I test.

SCBionicle:
I finally got it deployed. Am about to begin testing it. Didn't realize you had already merged.

itzg:
@SCBionicle I went ahead and did some testing since I got worried about parking a major refactor too long. Sorry, I should have communicated my timeline to you. 

Please do proceed and I'll jump on any fixes that are needed and/or glad to accept any fixes you would like to apply.

itzg:
BTW I had updated the examples further with docker-socket-proxy since a few more rules needs to be enabled.

SCBionicle:
> BTW I had updated the examples further with docker-socket-proxy since a few more rules needs to be enabled.

Yeah, figured that out the hard way. Needed to add some perms to the proxy before it would stop crashing. I opened a PR to add to the documentation, seemed you covered most of it, but there is a problem on multi node clusters if you don't specify the deployment constraints.

itzg:
Out of curiosity, what are the logs etc when it is deployed without the constraint? It seems like best practice would be to avoid deploying workloads on the master node.

SCBionicle:
```bash
mc_router.1.lwsbljbrt2ny@queen-bee    | time="2026-08-07T15:34:36Z" level=error msg="Docker failed to list services" error="Error response from daemon: This node is not a swarm manager. Worker nodes can't be used to view or modify cluster state. Please run this command on a manager node or promote the current node to a manager."
mc_router.1.lwsbljbrt2ny@queen-bee    | time="2026-08-07T15:34:36Z" level=fatal msg="Could not setup server" error="could not start docker swarm integration: Error response from daemon: This node is not a swarm manager. Worker nodes can't be used to view or modify cluster state. Please run this command on a manager node or promote the current node to a manager."
```

SCBionicle:
The thing with testing this on Docker Desktop is that initializing the Swarm cluster, you only have a manager to test on, there are no worker nodes. But simple rule of thumb, if you need to know the cluster state or modify the cluster state, the service that is doing the read or write must be a manager node.

SCBionicle:
I'm noticing an issue where it's not scaling down on swarm with RC1, after resolving issues with the socket proxy. I'm testing it on the latest build and seeing if the bug is still there before I post an issue.

From observations, it seems to not be firing the auto scale down timer. Unknown whether the timer is started or failed to fire when finished, logs do not indicate that other than the scale down event does not fire.

itzg:
RC1 for sure had bugs. I found several while testing docker swarm. Please use the latest release.

SCBionicle:
Latest did fix the scale down issue, so no issue.

PR #589, opened 2026-08-06
nil pointer error modifying routes via api without config loader

### LINKED ISSUE #588: Error modifying routes via API when mc-router runs without config file
Hi,

there is a bug in the routes api handlers (create, delete, set), that reveals itself when mc-router runs without a config file.
The `RoutesConfigLoader` instance on the server is `nil` in this case, but the handlers do not handle this case.
I have prepared a fix in a small PR.

moritzrinow:
Fixes #588


itzg (approved):
Thanks!

[Github is having an incident right now](https://www.githubstatus.com/incidents/qcvjkzcs7j74), so I'm going to wait on merging.

PR #590, opened 2026-08-07
Specify deployment constraints in compose file for docker swarm example

SCBionicle:
Missed a gotcha with Swarm that will lead to instant crashing unless this is specified.

### DIFF
--- a/examples/swarm/compose.yml
+++ b/examples/swarm/compose.yml
@@ -11,6 +11,10 @@ services:
       TASKS: 1
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock
+    deploy:
+      placement:
+        constraints:
+          - node.role == manager
     networks:
       - minecraft
 


itzg (approved):
I'm happy to merge this, but curious if this change is generally applicable? I use this compose file for my testing but had only setup a single node cluster.

SCBionicle:
This for the cases of a multi node cluster. You can't access docker swarm cluster states, nor modify them on nonmanager nodes. Single node clusters, the only node is both a worker and manager. (In Swarm, there is no such thing as manager only nodes)

itzg:
I missed that you put it on the socket proxy. Totally makes sense now

SCBionicle:
It's all good. 

Btw, very small thing that I usually practice on my swarm cluster since the ip range of an overlay network is very small (255 IP addresses, including both the task ip and the VIP). I typically make multiple overlay networks for different roles. In this case, the overlay network to the socket proxy is separate from the overlay network to the Minecraft server. If the Minecraft server is ever compromised, it cannot access the socket proxy directly.

itzg:
Great idea. Would be great if you could update the example with that.

PR #597, opened 2026-08-17
Fix delay of predefined (asleep/loading) MOTD serving for the Java client

### LINKED ISSUE #596: Asleep/loading MOTD never shows in the Minecraft client (server list just shows unreachable)
Running mc-router **v1.46.2** with Docker auto-scaling. When a backend is scaled down or waking up, the router serves the `auto-scale-asleep-motd` / `auto-scale-loading-motd` fine, but only if you query it with something like `mcstatus` or a python ping script. In the actual Minecraft client (26.2), the server list entry just shows "can't connect" / no MOTD at all, even though the server is listed with the right address and the router is clearly answering.

Ive been fighting this for ~4h now. When the client pings a scaled-down server, the router logs show the predefined status response IS being served, but only afer a 3 second timeout:

```
level=debug msg="Missing backend: serving predefined status response" server=vanilla.example.com
level=warning msg="Predefined status: error reading initial status packet" error="i/o timeout"
level=debug msg="Read frame" frame="Frame:[len=9, payload=0X0100001E4096EC0244]"
level=debug msg="Predefined status: no ping received, closing"
```

so the status response is written only after the 3s read deadline expires, and the ping / pong never completes, the client gets the response but never a pong, so it treats the whole ping as failed and shows nothing.

## Reproduction

1. mc-router with `IN_DOCKER=true`, `AUTO_SCALE_UP=true`, `AUTO_SCALE_DOWN=true`, and per-container labels `mc-router.auto-scale-asleep-motd` / `mc-router.auto-scale-loading-motd` (also happens with the global `AUTO_SCALE_ASLEEP_MOTD` / `AUTO_SCALE_LOADING_MOTD` env vars).
2. let a backend scale down (or catch it during wake-up).
3. ping the hostname from the Minecraft server list -> nothing shown.
4. ping it with any script that doesn't require the pong -> MOTD arrives, but with a ~3s delay.

### `docker-compose.yml` (my exact example)

```yaml
services:
  router:
    image: mc-router:patched
    container_name: mc_router
    ports:
      - "25565:25565"
    group_add:
      - "0"
    environment:
      IN_DOCKER: true
      AUTO_SCALE_UP: true
      AUTO_SCALE_DOWN: true
      AUTO_SCALE_DOWN_AFTER: "30s"
      AUTO_SCALE_ASLEEP_MOTD: "§4Sleeping"
      AUTO_SCALE_LOADING_MOTD: "§eLoading"
      DEBUG: true
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock:ro

  forge:
    image: itzg/minecraft-server
    container_name: mc_forge
    volumes:
      - "./forge:/data"
    environment:
      EULA: true
      TYPE: "FORGE"
      VERSION: "26.2"
      MOTD: |-
        §7Server Status: §2Online
        §7Server Version: §f%TYPE%  §7§o%VERSION%
    labels:
      mc-router.host: "forge.mcproxy.lokal"
      mc-router.auto-scale-up: true
      mc-router.auto-scale-down: true
      mc-router.auto-scale-asleep-motd: "§7Server Status: §4Sleeping\n§7Server Version: §fFORGE §8§o26.2"
      mc-router.auto-scale-loading-motd: "§7Server Status: §eLoading\n§7Server Version: §fFORGE §8§o26.2"

  vanilla:
    image: itzg/minecraft-server
    container_name: mc_vanilla
    volumes:
      - "./vanilla:/data"
    environment:
      EULA: true
      VERSION: "26.2"
      MOTD: |-
        §7Server Status: §2Online
        §7Server Version: §f%TYPE%  §7§o%VERSION%
    labels:
      mc-router.host: "vanilla.mcproxy.lokal"
      mc-router.auto-scale-up: true
      mc-router.auto-scale-down: true
      mc-router.auto-scale-asleep-motd: "§7Server Status: §4Sleeping\n§7Server Version: §fVANILLA §8§o26.2"
      mc-router.auto-scale-loading-motd: "§7Server Status: §eLoading\n§7Server Version: §fVANILLA §8§o26.2"
```

## Expected behavior

The predefined MOTD should be served immediately (the client sends the status request with the handshake) and the router should answer the ping request with a pong so the client actually displays it.

## Environment

- mc-router `v1.46.2` (Docker image `itzg/mc-router:latest` as of Aug 2026)
- Docker auto-scale up/down with docker-compose, no Kubernetes
- Client: Minecraft Java 1.21.x

I suspect the issue is in the predefined-status handling in `server/connector.go` (`serveStatus` / `findAndConnectBackend`): the clients status request gets consumed into the buffered handshake reader and is lost, so the router waits out its 3s read deadline; and the ping payload is stored as `*PingPayload` but asserted as a value, so the pong is never sent.

_I am working on a pull request._

jadebi:
# Fix predefined (asleep/loading) MOTD serving for the Java client

Fixes [#596](https://github.com/itzg/mc-router/issues/596)

## Problem

The auto-scale asleep/loading MOTDs served by mc-router never display in the actual Minecraft Java client, the server list shows the host as unreachable, even though tools like `mcstatus`/custom pingers receive the MOTD. Two bugs cause this:

### 1. Status request is swallowed by the handshake buffering -> 3s response delay

`HandleConnection` reads the handshake through a `bufio.Reader` wrapping a `TeeReader` that coies everything read into `inspectionBuffer`. The clients status request (0x00), typically sent in the same TCP segment as the handshake, ends up inside the bufio buffer / `inspectionBuffer` (no longer on the socket)

WHen serving a predefined status response, `findAndConnectBackend` creates a fresh `bufio.NewReader(frontendConn)`, so the already-buffered status request is lost. `serveStatus` blocks on its 3s read deadline before writing the response

### 2. Pong is never sent -> Java client fails the ping

`mcproto.ReadPacket` stores the ping payload as `*PingPayload`, but `serveStatus` asserts it with `.(mcproto.PingPayload)` _(value type)_. The assertion always fails, so `pingPending` stays false and the pong (0x01) is never written. The Java client requires a matching pong; without it, the entire server-list ping is treated as failed and no MOTD is displayed

Both predefined-status paths (`Missing backend` and `Scalable backend unreachable`) are affected, as is the loading MOTD, since they all go through `serveStatus`

## Fix

- pass the handshake frame length from `HandleConnection` into `findAndConnectBackend`
- in both predefined-status serving branches skip the handshake frame in the pre-read `inspectionBuffer` and chain the remaining buffered content with the live connection:

```go
_, _ = io.CopyN(io.Discard, preReadContent, int64(handshakeBytes))
br := bufio.NewReader(io.MultiReader(preReadContent, frontendConn))
```

this lets `serveStatus` read the status request immediately instead of waiting out the 3s deadline

- FIx both ping payload assertions in `serveStatus` to match the actual stored type: `.(*mcproto.PingPayload)`

### DIFF
--- a/server/connector.go
+++ b/server/connector.go
@@ -373,7 +373,7 @@ func (c *Connector) HandleConnection(frontendConn net.Conn) {
 				Debug("Got user info")
 		}
 
-		c.findAndConnectBackend(frontendConn, clientAddr, inspectionBuffer, handshake.ServerAddress, playerInfo, handshake.NextState, false, int(handshake.ProtocolVersion))
+		c.findAndConnectBackend(frontendConn, clientAddr, inspectionBuffer, packet.Length, handshake.ServerAddress, playerInfo, handshake.NextState, false, int(handshake.ProtocolVersion))
 
 	} else if packet.PacketID == mcproto.PacketIdLegacyServerListPing {
 		handshake, ok := packet.Data.(*mcproto.LegacyServerListPing)
@@ -393,7 +393,7 @@ func (c *Connector) HandleConnection(frontendConn net.Conn) {
 
 		serverAddress := handshake.ServerAddress
 
-		c.findAndConnectBackend(frontendConn, clientAddr, inspectionBuffer, serverAddress, nil, mcproto.StateStatus, true, 0)
+		c.findAndConnectBackend(frontendConn, clientAddr, inspectionBuffer, 0, serverAddress, nil, mcproto.StateStatus, true, 0)
 	} else {
 		logrus.
 			WithField("client", clientAddr).
@@ -423,7 +423,7 @@ func (c *Connector) serveStatus(frontendConn net.Conn, reader *bufio.Reader, ser
 	var pingVal int64
 	if err == nil && firstPkt != nil {
 		if firstPkt.PacketID == mcproto.PacketIdPingRequest {
-			if payload, ok := firstPkt.Data.(mcproto.PingPayload); ok {
+			if payload, ok := firstPkt.Data.(*mcproto.PingPayload); ok {
 				pingPending = true
 				pingVal = payload.Timestamp
 				logrus.WithFields(logrus.Fields{
@@ -461,7 +461,7 @@ func (c *Connector) serveStatus(frontendConn net.Conn, reader *bufio.Reader, ser
 		_ = frontendConn.SetReadDeadline(time.Now().Add(2 * time.Second))
 		if nextPkt, err2 := mcproto.ReadPacket(reader, frontendConn.RemoteAddr(), mcproto.StateStatus); err2 == nil && nextPkt != nil {
 			if nextPkt.PacketID == mcproto.PacketIdPingRequest {
-				if payload, ok := nextPkt.Data.(mcproto.PingPayload); ok {
+				if payload, ok := nextPkt.Data.(*mcproto.PingPayload); ok {
 					pingPending = true
 					pingVal = payload.Timestamp
 					logrus.WithFields(logrus.Fields{
@@ -598,7 +598,7 @@ func (c *Connector) cleanupBackendConnection(
 }
 
 func (c *Connector) findAndConnectBackend(frontendConn net.Conn,
-	clientAddr net.Addr, preReadContent io.Reader, serverAddress string, playerInfo *PlayerInfo, nextState mcproto.State, isLegacy bool, clientProtocol int) {
+	clientAddr net.Addr, preReadContent io.Reader, handshakeBytes int, serverAddress string, playerInfo *PlayerInfo, nextState mcproto.State, isLegacy bool, clientProtocol int) {
 
 	backendHostPort, resolvedHost, scalingTarget, waker, _ := c.routes.FindBackendForServerAddress(c.ctx, serverAddress)
 	cleanupMetrics := false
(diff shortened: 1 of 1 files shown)


itzg (commented):
I'll need to review these changes closely, so it might take a while. The MOTD handling worked fine for me during my testing of latest scaling changes, so I'll need to carefully re-read your analysis to see why the result was different for you.

itzg (commented):
Changes look good. I'll give it a quick test tomorrow especially since I could have sworn server list in latest client showed the appropriate MOTD.

jadebi:
I just discovered that the MOTD **DOES** get shown but only after those ~3s of sleeping.
It seems the "not showing at all" is a extra artifact by a mod I had installed.
It still takes longer than nesesary for the proxy to return the MOTD which causes weird behaviours with some client side mods.
The PR fixes the ~3s delay before MOTD return. 

I also forgot to mention when the MOTD return delay happens:
- Server asleep
- Server starting

When the Server is Online and ready the MOTD gets returned instantly as expected.

itzg:
Ok, yeah, that's the behavior I was seeing in the client's server list view. 

I'll need to look back at the original PR, because I'm now thinking that delay was intentional to give a server a moment to respond in case it was awake but overloaded.

itzg:
[This](https://github.com/itzg/mc-router/pull/488) was the original PR, but it doesn't seem like the delay was intentional.

itzg (approved):
I finally a chance to try it out. Very nice! And what a subtle bug.

PR #598, opened 2026-08-17
Avoid rearming scale-down timer on route reload

### LINKED ISSUE #594: Watched route reload scales down a backend with an active player
### Symptom

With watched routes and webhook autoscaling configured, rewriting or reloading the routes file while a player is connected can scale down that backend after `AUTO_SCALE_DOWN_AFTER`. The backend stops and the active player is disconnected.

The player remains connected for the entire delay. There is no disconnect event before mc-router logs `Scaling down backend server` and calls the sleeper.

**Expected:** reloading an unchanged route must not make an active backend eligible for scale-down. Scale-down should only become eligible after the last connection to that backend closes.

### Observed timeline

This occurred on mc-router v1.46.2. The scale-down happened exactly ten minutes after the route reload, while the same player connection was still open:

```text
06:10:39 Connecting to backend backendHostPort="vanilla:25565" player=<player>/<uuid>
06:10:59 Loading routes config file
06:10:59 Created route mapping backend="vanilla:25565" serverAddress=<server>
06:20:59 Scaling down backend server scalingTarget="vanilla:25565"
06:21:10 Finished relay backend->frontend player=<player>/<uuid>
06:21:10 Error observed on connection relay error="... connection reset by peer"
06:21:10 Closed connection to backend connectionCount=0 player=<player>/<uuid>
```

The connection count did not reach zero until after the backend had been scaled down.

### Reproduction

1. Run mc-router v1.46.2 with a watched routes file and a webhook sleeper.
2. Set a short `AUTO_SCALE_DOWN_AFTER`.
3. Connect a player through the router and keep the connection open.
4. Rewrite the routes file with the same non-empty mapping.
5. Wait past `AUTO_SCALE_DOWN_AFTER`.
6. Observe the scale-down callback fire and the active connection get reset.

Minimal relevant configuration:

```yaml
environment:
  ROUTES_CONFIG: /data/routes.json
  ROUTES_CONFIG_WATCH: "true"
  AUTO_SCALE_DOWN_AFTER: 10m
  AUTO_SCALE_WEBHOOK_URL: http://controller:8091/servers/autoscale
```

### Failing regression test

I reproduced the sequence from the v1.46.2 tag with an integration-style test. It opens a real proxied Minecraft login connection, verifies that mc-router counts it as active, re-registers the unchanged route as the watched-file loader does, and records the active count inside the sleeper.

<details>
<summary>Test case</summary>

```go
package server

import (
    "context"
    "io"
    "net"
    "testing"
    "time"

    "github.com/itzg/mc-router/mcproto"
    "github.com/stretchr/testify/require"
)

func TestRouteConfigReloadDoesNotScaleDownActiveConnection(t *testing.T) {
    const scaleDownDelay = 500 * time.Millisecond
    const serverAddress = "mc.example.com"

    downScaler := NewDownScaler(true, scaleDownDelay)
    routes := NewRoutes(t.Context()).WithDownScaler(downScaler)
    target := TestingScalingTarget("reload-test-backend")
    var connector *Connector
    scaleDownCalled := make(chan int, 1)
    sleeper := func(context.Context) error {
        scaleDownCalled <- connector.scaleActiveConnections.GetCount(target.ScalingKey())
        return nil
    }

    backendListener, err := net.Listen("tcp", "127.0.0.1:0")
    require.NoError(t, err)
    t.Cleanup(func() { _ = backendListener.Close() })

    backendConnection := make(chan net.Conn, 1)
    go func() {
        conn, acceptErr := backendListener.Accept()
        if acceptErr == nil {
            backendConnection <- conn
        }
    }()

    backendAddress := backendListener.Addr().String()
    routes.CreateMapping(serverAddress, backendAddress, target, nil, sleeper, "", "")

    connector = NewConnector(
        t.Context(),
        routes,
        downScaler,
        discardMetricsBuilder{}.BuildConnectorMetrics(),
        false,
        false,
        nil,
    )

    frontendListener, err := net.Listen("tcp", "127.0.0.1:0")
    require.NoError(t, err)
    t.Cleanup(func() { _ = frontendListener.Close() })
    go connector.acceptConnections(frontendListener, 100, 0)

    clientConnection, err := net.Dial("tcp", frontendListener.Addr().String())
    require.NoError(t, err)
    t.Cleanup(func() { _ = clientConnection.Close() })

    require.NoError(t, writeTestPacket(clientConnection, 0x00, func(w io.Writer) {
        _ = mcproto.WriteVarInt(w, int32(mcproto.ProtocolVersion1_18_2))
        _ = mcproto.WriteString(w, serverAddress)
        _, _ = w.Write([]byte{0x63, 0xdd})
        _ = mcproto.WriteVarInt(w, int32(mcproto.StateLogin))
    }))
    require.NoError(t, writeTestPacket(clientConnection, 0x00, func(w io.Writer) {
        _ = mcproto.WriteString(w, "RouteReloadTest")
    }))

    var acceptedBackendConnection net.Conn
    select {
    case acceptedBackendConnection = <-backendConnection:
        t.Cleanup(func() { _ = acceptedBackendConnection.Close() })
    case <-time.After(2 * time.Second):
        t.Fatal("router did not connect to test backend")
    }

    require.Eventually(t, func() bool {
        return connector.scaleActiveConnections.GetCount(target.ScalingKey()) == 1
    }, 2*time.Second, 10*time.Millisecond, "test connection never became active")

    // RoutesConfigLoader.Load calls BulkRegister after a watched file change,
    // which calls CreateMapping again for this unchanged, already-active route.
    routes.CreateMapping(serverAddress, backendAddress, target, nil, sleeper, "", "")

    select {
    case activeConnections := <-scaleDownCalled:
        t.Fatalf("scale-down sleeper ran with %d active backend connection(s)", activeConnections)
    case <-time.After(2 * scaleDownDelay):
    }
}
```

</details>

Result on v1.46.2:

```text
=== RUN   TestRouteConfigReloadDoesNotScaleDownActiveConnection
Created route mapping
Connecting to backend player=RouteReloadTest
Created route mapping
Scaling down backend server scalingTarget=reload-test-backend
routes_config_reload_test.go:90: scale-down sleeper ran with 1 active backend connection(s)
--- FAIL: TestRouteConfigReloadDoesNotScaleDownActiveConnection (0.50s)
```

This confirms that the sleeper is invoked while mc-router itself still reports one active backend connection.

### Implementation clue

The watched-file loader calls `Load`, which calls `BulkRegister`; that re-enters `CreateMapping` for the unchanged route. `CreateMapping` starts a down timer whenever the mapping has a sleeper. The existing connection cancelled the earlier timer when it connected, but the reload creates a later timer and there is no new connection event to cancel it.

- Watched reload: https://github.com/itzg/mc-router/blob/v1.46.2/server/routes_config_loader.go#L119-L130
- Route registration: https://github.com/itzg/mc-router/blob/v1.46.2/server/routes.go#L369-L390
- Connection count and timer cancellation: https://github.com/itzg/mc-router/blob/v1.46.2/server/connector.go#L765-L775
- Timer callback: https://github.com/itzg/mc-router/blob/v1.46.2/server/down_scaler.go#L69-L90

The likely defect is therefore not that a timer exists, but that route re-registration starts a new one for an already-active backend and its callback does not revalidate the active connection count.

Buzzec:
## What changed

- preserve scale-down timer state when an already-routable scaling target is re-registered
- add an end-to-end regression test with a real proxied Minecraft login connection

## Why

Watched routes are reloaded through `CreateMapping`. That method always started a scale-down timer, even when the same scaling target was already routable. If an active connection had previously cancelled the timer, reloading the routes file silently re-armed it and allowed the backend to be stopped beneath the connected player.

The timer now starts only when a mapping first makes a scaling target routable, when the previous mapping was down, or when the scaling target changes.

## Impact

Reloading a watched routes file no longer makes an actively used backend eligible for scale-down. Initial registrations and down-to-ready transitions continue to receive their idle timer.

## Validation

- `go test ./...`
- `go test -race ./server`
- `go test ./server -run '^TestRouteConfigReloadDoesNotScaleDownActiveConnection$' -count=10`

Fixes #594

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -376,16 +376,19 @@ func (r *routesImpl) CreateMapping(serverAddress string, backend string, scaling
 		"serverAddress": serverAddress,
 		"backend":       backend,
 	}).Info("Created route mapping")
+	previous, hasPrevious := r.mappings[serverAddress]
 	r.mappings[serverAddress] = mapping{backend: backend, scalingTarget: scalingTarget, waker: waker, sleeper: sleeper, asleepMOTD: asleepMOTD, loadingMOTD: loadingMOTD}
 
 	for _, listener := range r.routesListeners {
 		listener.OnRouteAdded(serverAddress, backend)
 	}
 
-	// Trigger auto-scale down when mapping is created to ensure servers are shut down if router restarts.
-	// Only start the timer when backend is non-empty — an empty backend means the server is already
-	// asleep/stopped (e.g. a Docker stop event updated the route), so there is nothing to scale down.
-	if r.downScaler != nil && scalingTarget != nil && backend != "" {
+	// Trigger auto-scale down when a mapping first makes a scaling target routable to ensure servers
+	// are shut down if the router restarts. Re-registering an already-routable target must preserve
+	// its timer state since an active connection may have cancelled the timer in the meantime.
+	shouldStartDownTimer := scalingTarget != nil && (!hasPrevious || previous.backend == "" || previous.scalingTarget == nil ||
+		previous.scalingTarget.ScalingKey() != scalingTarget.ScalingKey())
+	if r.downScaler != nil && backend != "" && shouldStartDownTimer {
 		r.downScaler.Start(r.ctx, scalingTarget, r)
 	}
 }
--- a/server/routes_config_reload_test.go
+++ b/server/routes_config_reload_test.go
@@ -1,0 +1,93 @@
+package server
+
+import (
+	"context"
+	"io"
+	"net"
+	"testing"
+	"time"
+
+	"github.com/itzg/mc-router/mcproto"
+	"github.com/stretchr/testify/require"
+)
+
+func TestRouteConfigReloadDoesNotScaleDownActiveConnection(t *testing.T) {
+	const scaleDownDelay = 500 * time.Millisecond
+	const serverAddress = "mc.example.com"
+
+	downScaler := NewDownScaler(true, scaleDownDelay)
+	routes := NewRoutes(t.Context()).WithDownScaler(downScaler)
+	target := TestingScalingTarget("reload-test-backend")
+	var connector *Connector
+	scaleDownCalled := make(chan int, 1)
+	sleeper := func(context.Context) error {
+		scaleDownCalled <- connector.scaleActiveConnections.GetCount(target.ScalingKey())
+		return nil
+	}
+
+	backendListener, err := net.Listen("tcp", "127.0.0.1:0")
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = backendListener.Close() })
+
+	backendConnection := make(chan net.Conn, 1)
+	go func() {
+		conn, acceptErr := backendListener.Accept()
+		if acceptErr == nil {
+			backendConnection <- conn
+		}
+	}()
+
+	backendAddress := backendListener.Addr().String()
+	routes.CreateMapping(serverAddress, backendAddress, target, nil, sleeper, "", "")
+
+	connector = NewConnector(
+		t.Context(),
+		routes,
+		downScaler,
+		discardMetricsBuilder{}.BuildConnectorMetrics(),
+		false,
+		false,
+		nil,
+	)
+
+	frontendListener, err := net.Listen("tcp", "127.0.0.1:0")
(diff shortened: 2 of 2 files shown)


itzg (approved):
Thanks!

PR #614, opened 2026-09-04
Abstract ConnectionMetrics to allow for provider switch

itzg:
go-kit is a bloated set of depenencies and no longer the prevailing way to expose metrics, such as for Prometheus. As such, this PR is the first step abstract the increment/set/add operations behind methods.

The goal is to switch to go.opentelemetry.io/otel with go.opentelemetry.io/otel/exporters/prometheus.

InfluxDB will probably be the trickier one since there isn't a otel exporter for that. Ideally an otel collector exporter option would allow users to backfill that.


itzg:
Includes deprecation notice for https://github.com/itzg/mc-router/issues/615

PR #617, opened 2026-09-16
Add dynamic proxy protocol

nelsongraca:
Added dynamic proxy protocol support.

Allows to accept connections with or without PROXY protocol, if a connection supplies the PROXY header it is forwarded to the backend verbatim, if no PROXY protocol is present nothing is forwarded.

It can't be used together with -receive-proxy-protocol as it would be redundant, can't also be used with -use-proxy-protocol, no specific reason just because makes sense in my opinion

How it works:

Internally it sets `receiveProxyProto` to create a proxy protocol listener, this causes the listener policy to be set to `USE` (like it was already with `-receive-proxy-protocol`), this makes the PROXY header optional on incoming connections, allowing to receive both.
On `findAndConnectBackend` if dynamic proxy protocol is enabled the PROXY header received on the connection is directly forwarded to the backend.


itzg:
Since you didn’t explain how the changes ended up being so simple, I thought I would ask Gemini. And I wanted future-me to know how this worked 😄 

- - -

The magic happens by combining the go-proxyproto library's connection wrapping with Go's type assertion.
Here is a breakdown of how it works under the hood:
 * **Enabling Proxy Detection:** When UseDynamicProxyProtocol is enabled, c.receiveProxyProto is set to true. This causes incoming listener connections to be wrapped using the go-proxyproto library listener/connection wrapper.
 * **Header Parsing:** When a client connects, the go-proxyproto listener automatically checks the incoming stream for a valid PROXY protocol header:
   * If a PROXY header is present, it parses it and wraps the underlying connection in a *proxyproto.Conn type.
   * If no PROXY header is sent, it leaves the connection as a standard raw TCP connection.
 * **Dynamic Header Extraction:** In server/connector.go, the new frontendProxyHeader helper function inspects the connection using Go's type assertion (conn.(*proxyproto.Conn)):
   ```go
   func frontendProxyHeader(conn net.Conn) *proxyproto.Header {
       if proxyConn, ok := conn.(*proxyproto.Conn); ok {
           return proxyConn.ProxyHeader()
       }
       return nil
   }
   
   ```
 * **Conditional Forwarding:** When connecting to the backend server in findAndConnectBackend:
   * If dynamicProxyProto is active, it calls frontendProxyHeader(frontendConn).
   * If a header was parsed (header != nil), it writes that header to backendConn.
   * If no header was present (header == nil), it skips sending a PROXY header entirely and proceeds with a standard connection.

itzg:
Let me know when you’re ready for a final review. It seems good so far.

nelsongraca:
> Since you didn’t explain how the changes ended up being so simple

Sorry 😅 I'll add the explanation, I think I can do a better one. Did some testing and looked good, but I want to inspect the actual traffic to be sure.

nelsongraca:
Tested, seems to be working fine

itzg:
I already like Gemini's explanation. The missing bit for me was that the proxy library was able to handle non-proxy detection. The rest makes total sense.

itzg (approved):
Thanks for requesting AND contributing the feature!

itzg:
Just noting that we discussed this [in Discord](https://discord.com/channels/660567679458869252/909522789290110978/1549872941867597825).

PR #620, opened 2026-09-18
Fix Docker Swarm client host discovery

### LINKED ISSUE #567: Docker Swarm Watcher ignores default Docker socket and standard environment variables
## Description
When running in Docker Swarm mode (`-in-docker-swarm`), the application fails to auto-discover the Docker daemon socket or respect standard Docker environment variables (such as `DOCKER_HOST`), requiring the user to explicitly specify the socket path via flags or manual configuration.

This differs from the standard Docker integration (`-in-docker`), which falls back gracefully to standard defaults.

## Root Cause
The bug resides in [docker_swarm.go](server/docker_swarm.go) during Docker client initialization inside `Start()`:

1. **Missing `client.FromEnv`**:
   Unlike [docker.go](server/docker.go), the Swarm watcher implementation does not include the `client.FromEnv` option when creating the client. This prevents the Docker SDK from loading configuration parameters from environment variables.

2. **Unconditional `client.WithHost("")`**:
   The client initialization calls `client.WithHost(w.config.socket)` unconditionally:
   ```go
   opts := []client.Opt{
       client.WithHost(w.config.socket),
       client.WithTimeout(w.config.timeout),
       client.WithHTTPHeaders(map[string]string{
           "User-Agent": "mc-router ",
       }),
       client.WithAPIVersionNegotiation(),
   }
   ```
   If the `-docker-socket` flag is not specified, `w.config.socket` defaults to `""`. This forces `client.WithHost("")` which overrides the standard SDK defaults (e.g. `unix:///var/run/docker.sock` on Unix or `npipe:////./pipe/docker_engine` on Windows) to an empty string.

## Proposed Fix
Align the Swarm watcher initialization with the standard Docker watcher implementation in [docker.go](server/docker.go):

1. Add `client.FromEnv` to the `opts` slice.
2. Only append `client.WithHost(w.config.socket)` if `w.config.socket` is non-empty.

```diff
-	opts := []client.Opt{
-		client.WithHost(w.config.socket),
-		client.WithTimeout(w.config.timeout),
-		client.WithHTTPHeaders(map[string]string{
-			"User-Agent": "mc-router ",
-		}),
-		client.WithAPIVersionNegotiation(),
-	}
+	opts := []client.Opt{
+		client.FromEnv,
+		client.WithTimeout(w.config.timeout),
+		client.WithHTTPHeaders(map[string]string{
+			"User-Agent": "mc-router ",
+		}),
+		client.WithAPIVersionNegotiation(),
+	}
+
+	if w.config.socket != "" {
+		opts = append(opts, client.WithHost(w.config.socket))
+	}
```

---

**AI Disclosure**: The above was investigated and drafted with AI.

**Context**: I run a multinode Docker Swarm cluster in my homelab. I ran into this issue when setting up my Minecraft server behind MC-Router. I couldn't find any documentation on why the Router was crashing, so I tasked an agent to find out why and suggest a fix both for my Swarm configuration and this issue.
**My Current Work Around**: Set the following in MC-Router's environment configuration `DOCKER_SOCKET: "unix:///var/run/docker.sock"`

HP-network:
The Swarm watcher did not load Docker client settings from the environment and always applied its socket option. When no socket flag was provided, the empty host replaced the Docker SDK default and ignored DOCKER_HOST.

Use the same client option builder for both Docker watchers. Environment values are loaded first, while an explicitly configured socket still takes precedence. The configured API version is also honored by the Swarm watcher.

Tests: go test ./...

Closes #567.


itzg (approved):
Thanks! I'll merge and release this in the morning.

